PR #24, opened 2026-08-04
configd, and a BLE client to drive it

pierre-rouanet:
Stacks on #21 (`btd`). Adds the service `btd` was missing, and the laptop-side client that makes the whole path testable.

## `configd` — wifi and the robot's identity

`net.status`, `net.scan`, `net.connect`, `net.forget`, `system.info`, `system.setName`, `system.reboot`, over a fifth unix socket.

A separate service rather than part of `robotd`, because **config must be reachable when the robot is dead** — provisioning wifi is exactly what a client needs when things are broken (§3.1). And not part of `btd`, because `btd` owns nothing (§4.1): an SDK should not have to go through Bluetooth to set a name.

**It stores no credentials.** NetworkManager owns those, persists them root-only and reconnects by itself. A passphrase passes through `configd` and is forgotten.

### The distinction that justified choosing NM

`ConnectFailure::BadKey`. "You typed the password wrong" is the most common provisioning failure there is, and netplan cannot report it — `netplan apply` says "config applied" whether or not association succeeded. NM reports it as device state reason 7, and the mapping is tested. That's the payoff from the earlier netplan-vs-NM comparison, now in the protocol.

### The passphrase never reaches a log

`NetConnectParams` has a hand-written `Debug` that redacts `psk`, with one test asserting the key never appears in `{:?}` and another asserting it still reaches the wire. Every other params struct derives `Debug`; a derived one here would put a customer's wifi password in the journal the first time any service logged a request it couldn't handle.

### Why it runs as root

Narrowly: NM's connection-modify and logind's `Reboot` are both polkit-gated, there is **no polkit on this image**, and systemd denies both to a session-less non-root caller. The alternative was installing a JS policy engine to authorise two calls.

The trust boundary stays right: `btd` — which parses bytes from anyone in radio range — is unprivileged, and `configd` only ever sees typed JSON over a peer-credentialled socket. Unlike `robotd` it touches no hardware, so the unit sandboxes it properly: `ProtectSystem=strict`, one writable path, `AF_UNIX` only, empty `CapabilityBoundingSet`. `CAP_SYS_BOOT` is deliberately absent — logind performs the reboot, `configd` only asks, and a capability there would permit the unclean `reboot(2)` this exists to avoid.

`system.reboot` answers **before** rebooting, after 3s. A daemon that rebooted inside the call would drop the connection before responding, and every client would have to treat a broken pipe as success.

### The store

A file plus `flock` and `rename(2)`, not a service (§3.1) — readable when anything is down, and the updater never touches it so it survives an update *and* a rollback. Names are trimmed, stripped of control characters, truncated to 24 chars on a character boundary; the stored name is returned so a client displays what the robot has rather than what it sent. A missing or corrupt file yields the hostname rather than refusing to start.

## `btd` routed all seven, and the exhaustive match earned itself

Adding the methods to the protocol **broke `btd`'s build** until someone decided about each one — which is exactly what that match is for. Provisioning is the case BLE exists for (a robot that has never seen a network cannot be configured over that network), so all seven are permitted, and the test naming which mutating calls BLE may make grew from one entry to five.

`updaterd` had the same protection and the same break; it now names `configd` for `net.*`/`system.*` as it already named `robotd` for `robot.*`.

## `btctl` — the client

```
cargo run -p btd --example btctl -- scan
cargo run -p btd --example btctl -- wifi scan
cargo run -p btd --example btctl -- wifi connect "Pollen" --psk secret
cargo run -p btd --example btctl -- name "Ducky"
cargo run -p btd --example btctl -- call robot.health
```

An **example, not a binary**, so `btleplug` never reaches the robot. `btleplug` rather than `bluer` because this half runs on a laptop — CoreBluetooth on macOS — and `bluer` would make the client Linux-only.

It reuses `btd::framing`, which is the point: the chunking here is the *client* half of the same module the robot uses, so asymmetric framing simply wouldn't work. That makes it a test of the protocol rather than a reimplementation free to agree with itself.

Writing it surfaced two real flaws:

- **The GATT UUIDs were in `bluez.rs`**, which is `cfg(target_os = "linux")` — so a client couldn't reach them off-Linux. They're a wire contract like the method names, so they move to `gatt.rs` ungated.
- **`FramingError` implemented neither `Display` nor `Error`**, so `?` didn't work on it. Fine in a binary, wrong for a library.

## Testing

Whole workspace passes, and **every binary cross-compiles for aarch64** (`cargo board --bins`), so the zbus proxies and the bluer backend both type-check for the target.

- proto: 19, including the two redaction tests
- `configd`: 14 against an in-memory wifi stack that produces a wrong-key failure on demand
- `btd`: 23
- `updater`/`robotctl`/`robotd`: unchanged and passing

**Never run against a real NetworkManager or a real radio.** `configd --fake-net` serves the whole `net.*` surface from memory, which is how to exercise it without a board.

## Not in here, and worth deciding

- **`robotctl net` / `robotctl system`.** The on-robot client for these methods. `btctl` covers the BLE path, but someone on the board still has no CLI for wifi.
- **BLE encryption and pairing.** Still the gap from #21, and now it matters more: `net.connect` carries a passphrase over an unencrypted characteristic. §7 requires pairing plus encryption, and the flag is one line — but it needs a decision about how a phone may bond (a button-held window, or a secret under the robot). **I would not put this on a customer robot until that lands.**
- **Docs.** `architecture.md` §1 should list `configd`, and `roadmap.md` should say M6's BLE work started early.
- **`on_apply`'s restart set** in `updater.toml` is still `["robotd"]`. Two new daemons need to join it or an update leaves the old ones running.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,126 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-broadcast"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "435a87a52755b8f27fcf321ac4f04b2802e337c8c4872923137471ec39c37532"
+dependencies = [
+ "event-listener",
+ "event-listener-strategy",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-channel"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "924ed96dd52d1b75e9c1a3e6275715fd320f5f9439fb5a4a11fa51f4221158d2"
+dependencies = [
+ "concurrent-queue",
+ "event-listener-strategy",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-executor"
+version = "1.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c96bf972d85afc50bf5ab8fe2d54d1586b4e0b46c97c50a0c9e71e2f7bcd812a"
+dependencies = [
+ "async-task",
+ "concurrent-queue",
+ "fastrand",
+ "futures-lite",
+ "pin-project-lite",
+ "slab",
+]
+
+[[package]]
+name = "async-io"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
+dependencies = [
+ "autocfg",
+ "cfg-if",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite",
+ "parking",
+ "polling",
+ "rustix",
+ "slab",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "async-lock"
+version = "3.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
+dependencies = [
+ "event-listener",
+ "event-listener-strategy",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-process"
+version = "2.5.0"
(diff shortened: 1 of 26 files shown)


pierre-rouanet:
Folded into #21 — everything is on `btd-ble-service` now, so there is one PR to review rather than a stack.

PR #26, opened 2026-08-04
robotctl health reports the whole robot

pierre-rouanet:
**Stacked on #23** — base is `battery-in-health`, so this diff shows only the delta. Merge #23 first and this retargets to `main` on its own.

`robotctl health` printed a verdict and a battery line. The verdict was both the useful part and the frustrating one: `unhealthy: control loop at 43.9 Hz` says what is wrong and nothing about why, and answering that meant a second command, the journal, or hand-written JSON at the socket.

```
$ robotctl health
robot     healthy
  loop      50.1 of 50.0 Hz · 2834 ticks · 0 missed · last 13 ms ago
  bus       ok
  imu       ready
  battery   7.62 V (64%)
  motors    41 °C max (left_knee) · 36 °C mean

software
  updaterd  0.1.4 (rev abc1234)
  robotd    0.1.5 (rev def5678)
  daemon    0.1.5 installed
            last update 0.1.4 → 0.1.5: applied
```

## One command, both halves

"What is wrong with this robot" does not divide into hardware and software until after it is answered — a robot that reverted a release an hour ago looks exactly like a robot with unpowered servos until both are on screen together. `robotctl version` stays the software half on its own, and the gathering is shared (`collect_version_report`) so the two commands cannot disagree.

## What `robot.health` now carries

The sections behind the verdict: `control_loop` (achieved against target, ticks, missed, last-tick age), `bus` (consecutive errors, startup failures), `imu` (ready, stale reads), `battery`, `motors`. All optional or defaulted, so an older `robotd` still parses.

The loop section carries the numbers the verdict was *computed from*, which is the point: 43.9 Hz next to `missed = 0` says the loop is being woken late, not doing too much. That distinction is what made the `Delay`→`Skip` fix in c5c1187 a one-liner instead of a hardware investigation.

## Motor temperature comes for free

`present_temperature` (146) is one register past `present_input_voltage` (144), so a 3-byte read gets both and thermals cost no extra bus transaction — `bus_voltage` becomes `slow_sensors` accordingly. Reported as the **hottest** joint, named: a knee holding a squat runs far above the mouth, and a mean over fifteen servos hides the one servo approaching the overheat shutdown its error mask latches on.

## Still reported, never judged

`healthy`/`degraded` come only from what a release can be blamed for. Battery, thermals and counters are descriptions, and tests hold that line — a verdict gating on temperature would roll good releases back on a warm afternoon, then judge their replacements on the same warm afternoon. §4.5 of `robotd-design.md` states the rule for whatever field is added next.

## Exit codes, and a dead robotd

Contract unchanged, since `install.sh` gates on it: `0` healthy, `REFUSED` unhealthy, `UNREACHABLE` when nothing answered. What changed is that an unreachable `robotd` no longer aborts the command — it prints the software half anyway, because "the robot does nothing" is very often a daemon that failed to start, and what is installed is then the half that explains it.

Two more things nobody thinks to ask about, both of which explain "updates stopped working": a **pinned** component, and the **last update attempt** with the reason behind an automatic revert — which outside the journal exists nowhere else.

## Testing

`cargo test --workspace` green (17 suites, 310 tests), `clippy --all-targets` and `fmt` clean. Verified by hand against a real `robotd` plus a real `updaterd` on the playground: healthy, `robotd` stopped (software half survives, exit 3), and `--json`.

New tests cover the cases a live robot never produces — nothing measured yet rendering as unknown rather than `0.0 Hz`/`0.00 V`, a stopped `robotd` still yielding software, a stale IMU, a bench board waiting on its bus, a pin, and a rollback summary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ the robot is aarch64 Linux.
 cargo test --workspace
 ```
 
-272 tests, no hardware, no network, no Docker. If they pass, your checkout is sound.
+310 tests, no hardware, no network, no Docker. If they pass, your checkout is sound.
 
 The fastest way to actually *see* the update engine work is the playground, which drives
 the real engine — real signatures, real atomic swaps, real rollback — against a fake remote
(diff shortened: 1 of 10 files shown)


pierre-rouanet:
Folded into #23 — same three commits now live on `battery-in-health`, so everything merges at once from there.

PR #32, opened 2026-08-05
provisioning: make a fresh board's own advice runnable

pierre-rouanet:
Bringing up a Radxa Zero 3W from a clean Armbian image hit four dead ends, all the same kind: a command printed by a script or by `deploy/README.md` that could not be pasted into the shell it was printed to.

### The token

Both README fetches omitted it. `raw.githubusercontent.com` answers **404**, not 401, for a private path with no credentials, so step one of provisioning fails as what looks like a wrong URL. `install.sh` already had a `DUCK_TOKEN` path; nothing told anyone to use it — and the documented `curl … | sudo sh` cannot carry it either way, since the header goes on the fetch and the variable has to cross `sudo` separately. There is now a step 0 that sets up `$AUTH`/`$RAW` once, and step 2 is two commands.

### `migrate-network.sh` was unreachable, then refused

`setup-board.sh` printed `sudo sh migrate-network.sh` — a bare relative name, for a sibling script a fresh board has never fetched. It now names `/tmp`, and prints the fetch command when the file is not there yet. The README gained it as step 1b, including the fact that it runs **twice**, either side of the reboot.

Once fetched, it refused: `/run/netplan/wpa-wlan0.conf` did not exist on this board, and the error told the operator to hand-type a key netplan already had in plain sight in `/etc/netplan/30-wifis-dhcp.yaml`. The runtime file is a *derived* artifact — the YAML is the source. It now falls back to the YAML, handling the `SSID: {}` spelling of an open network and netplan's `key-management: sae`. Refusing outright is still the behaviour when neither source can answer; a wrong guess there strands a headless board.

### The kernel-console warning contradicted the fix

It fired on the very run that fixed it — "set `console=display` and reboot", right after the script had done exactly that. `/proc/cmdline` is the kernel that is running; `armbianEnv.txt` is the one that will run. Three states now:

| | |
|---|---|
| file says `display`, reboot pending | `ttyS2 (the motor bus), until the reboot` — no warning |
| file says `display`, boot disagrees | `— CONFLICT`, and says an `extraargs=` or U-Boot `bootargs` wins, so editing `console=` again will not help |
| file says neither | the original warning, now stating that the script only rewrites `console=both` and `console=serial` |

`ttyFIQ*` counts as a serial console too, and the report names the tty it found rather than hedging with "if that is /dev/ttyS2".

### Keeping it from drifting back

`migrate-network.sh` was missing from the CI lint loop while being the riskiest of the three scripts — the one that can make a headless board unreachable, and the one nobody re-runs once their own board is migrated. It is in the loop now, and a new xtask test asserts that no printed command names a bare relative script path. It catches literals only, so it narrows that failure rather than closing it.

### Verified

- `shellcheck --shell=sh` and `sh -n` clean on all three scripts; the CI lint step reproduced locally.
- `cargo test -p xtask` green, and the new test confirmed to fail when the old bare-name advice is put back.
- The YAML parser exercised against six fixtures: quoted SSID with spaces, open network via `{}`, WPA3/`sae`, 64-hex PSK, a `wifis:` stanza with no `access-points:` (refuses), and a block followed by `dhcp4: true` (which must not be read as an SSID).
- The console logic exercised against fixture `cmdline`/`armbianEnv.txt` pairs for all three states plus `ttyFIQ0` and a clean board.

Not verified on hardware: the board is mid-bring-up. What was reproduced by hand there is the failure each change addresses.


pierre-rouanet:
Pushed a second commit: the first `robotctl health` on a correctly installed board reports a stopped daemon when nothing is stopped.

```
robot     unavailable — cannot reach robotd at /run/robotd.sock: Permission denied (os error 13)
! Is the service running?  systemctl status robotd
```

Both daemons were up. The sockets are `root:robot` mode 0660, `install.sh` does add the operator to `robot` — and a group only takes effect in a **new login session**, so the shell that ran the install is exactly the one that cannot connect. The advice then sends that person to look at two healthy units.

Every `io::ErrorKind` was getting the same hint, and it is the right question for one of them. `EACCES` from `connect` means something was found and refused us, never that it was absent (that is `ENOENT`), so the cases separate without stat'ing anything:

| kind | hint |
|---|---|
| `EACCES` | group membership, daemon is up — names the mode, the login-session rule, `id -nG`, and `sudo` as the interim |
| `ENOENT` | nothing listening there, and *is this even the socket it was told to use* — `--socket` and `robot_socket` both move it |
| `ECONNREFUSED` | a socket file outliving the process that made it |
| other | unchanged |

Every hint keeps the cause on its first line, because `health` renders that into a fixed-width column and pushes the rest into its warnings block — a hint that led with detail would be truncated into nonsense there. Three tests pin that, plus the two behaviours worth keeping: `EACCES` must not mention `systemctl`, and the other kinds must.

`deploy/README.md` now states the same thing where the `robot` group is described, since "the first health check after an install fails and the install is fine" is worth knowing before you see it.

`cargo test -p robotctl` (30) and `-p xtask` (3) green, clippy and fmt clean.

pierre-rouanet:
Third commit, answering two things from the board.

### "Can the group step be automatic?"

Not fully, and it is worth being precise about why: a process's group set is fixed at `exec`, and there is no API to add a group to a *running* process — not even for root. The shell that launched the installer can never be fixed from inside it. Future logins were already automatic (that is what `usermod -aG` did).

What can be shortened is the step that remains:

```
newgrp robot
```

One command in the same shell, instead of dropping an SSH session and coming back. `install.sh` now prints it in the closing report — gated on having actually run `usermod` this run, and placed **before** the command list, because every command in that list fails with `Permission denied` in the shell reading it. That is exactly what a first-time operator reads as a broken install. `robotctl`'s `EACCES` hint says the same thing, so the two cannot drift.

### `DUCK_DEV_KEY` was implemented and documented nowhere

It has been in `install.sh` since #18 — it validates the key, installs it under the load-bearing `team.dev.pub` name, and flips `allow_dev_keys` in one step. But every place explaining how to make a dev board still listed the two manual steps, and `docs/slice-2-bringup.md` still said *"once #18 is merged, install.sh does both steps"*. So the only documented route was the one where doing half of it leaves a board that refuses branch builds with a signature error that reads like a corrupt release.

```bash
scp ~/.duck-keys/team.dev.pub pierre@radxa-zero3:/tmp/
sudo DUCK_TOKEN="$DUCK_TOKEN" DUCK_DEV_KEY=/tmp/team.dev.pub sh /tmp/install.sh
```

`deploy/README.md` also gained the full knob table — `DUCK_TOKEN`, `DUCK_REPO`, `DUCK_REF`, `DUCK_DEV_KEY`, `DUCK_FORCE_REINSTALL` — which were not listed together anywhere, plus where `team.dev.pub` comes from given it is deliberately not in the repo. `README.md` leads with the `install.sh` route and keeps the by-hand steps for a board provisioned some other way.

Verified: `report()` rendered both ways (group pending and not), shellcheck/`sh -n` clean on all three scripts, `cargo test -p robotctl -p xtask` green, fmt and clippy clean.

pierre-rouanet:
Added `scripts/provision.sh`. **11 pasted blocks → 4**, and `newgrp robot` is gone entirely.

```bash
export DUCK_TOKEN=github_pat_...
curl -fsSL -H "Authorization: Bearer $DUCK_TOKEN" .../provision.sh -o /tmp/provision.sh && sudo DUCK_TOKEN="$DUCK_TOKEN" sh /tmp/provision.sh
sudo reboot
sudo /usr/local/sbin/robot-provision
```

It is glue and nothing else — the three scripts stay independently runnable and keep their separate lifetimes and risks. What they never had was anything tying them together, so the operator was the glue: fetch, run, fetch, run, reboot, re-run, re-run, fetch, run.

### The newgrp step is removed, not automated

Credit to Pierre for the question that got here. A script can never fix its parent shell's group set — but the group does not have to be created *late*. `install.sh` adds the operator last, after the reboot, so the shell reading its output predates the group's existence. Creating it in phase 1 means the reboot does that work: you reconnect, that is a new login session, and it already has `robot`.

`install.sh` still owns the decision and still installs the `sysusers.d` file from the verified release — it just finds the operator already a member and says so. Phase 1 needs nothing from the release, since `groupadd --system` is self-contained.

### Two things only a board would have shown

- **A dev key at `/tmp/team.dev.pub` is gone by phase 2**, because `/tmp` does not survive the reboot. Phase 1 copies it into `/var/lib/robot`. Without that, provisioning succeeds and silently produces a board that is *not* a dev board — surfacing much later as `--ref` being refused, which reads like a broken release.
- **"Have you rebooted" is the kernel's `boot_id`, not file mtime against uptime.** The mtime version was the first attempt and it is wrong on this hardware: no battery-backed RTC, so the clock starts at 1970 and NTP steps it *during* provisioning. Comparing timestamps would compare two clocks that disagree by decades. There is no clock in the check now.

### Deliberate choices

- **Always asks for a reboot** rather than deciding if one is needed — deciding means re-deriving what the two scripts already decided, or parsing their output. Thirty seconds buys phase 2 live boot config and a refreshed login session.
- **Refuses phase 2 in the same boot as phase 1**: installing onto a staged-but-not-live overlay gives `robotd` no `/dev/ttyS2`, and a hardware fault report that is really a missing step.
- **Does not take the reboot itself.** Both existing scripts state they never do, and the moment after a wifi cutover is when a headless board can go missing — that belongs to whoever can go and find it.
- **The token crosses the reboot in `/var/lib/robot/provision.env`**, 0600, root-only, deleted when provisioning finishes. This retires the `~/.profile` line from two commits ago, which put a credential in a file the operator keeps and that outlived the provisioning it was for.

Exercised against stubs for six paths: phase 1 with and without a token, the dev-key copy, the same-boot refusal, phase 2 through the persisted copies, and the state file's removal. Added to CI's shellcheck loop and to the absolute-paths test. `deploy/README.md` is down to 392 lines, the first 80 being the four commands twice.

Nothing here has run on hardware yet.

PR #44, opened 2026-08-07
Decision wanted: two ways apply is inert when it should not be

pierre-rouanet:
`already_current` on running revisions, and `apply --force`. Both were proposed
inside `install-path-gap.md` §4 while diagnosing an afternoon of `configd` updates
that reported success and changed nothing. Neither has been decided, and buried in a
post-mortem is the wrong place for a decision — so they move here, with what actually
needs settling rather than only what was proposed.

The document leads with the part that is easy to get wrong now: **neither is urgent**.
The restart set is derived from the release and `updaterd`/`btd` restart themselves
after replying, so the state these dig you out of is much harder to reach than when
they were written. They are still right; nothing is on fire.

What the write-up adds beyond restating the proposals:

- item 1's real question is not whether to compare running revisions — that is clearly
  right — but whether a mismatch should be **reported and refused** or **repaired**.
  Repairing makes `apply` idempotent-with-repair instead of idempotent-and-inert, and
  it largely dissolves item 2. Recommended, because the incident was expensive
  precisely because the recovery command was inert, and better wording on an inert
  command is not much of a fix;
- item 2 must not copy `install --force`'s guard by symmetry. That one refuses while
  `robotd` answers because it disables the health gate; `apply` keeps the gate, so the
  same guard would disable the flag exactly when a robot is up and skewed;
- both have a `btd`-shaped hole: `btd` serves no socket, so its running revision cannot
  be read at all. Any design here should say so rather than quietly skip it.

No code, deliberately. Both change what `apply` does on a board, and the last round of
correct-but-too-narrow fixes in this area is what `install-path-gap.md` exists to
record.

**No code in this PR by design.** It is a decision to make, opened as a draft so it is visible rather than buried in `install-path-gap.md` §4.

### DIFF
--- a/docs/recovery-when-apply-does-nothing.md
+++ b/docs/recovery-when-apply-does-nothing.md
@@ -1,0 +1,89 @@
+# When `apply` correctly does nothing and that is the wrong answer
+
+Status: **decision wanted, nothing implemented** · Date: 2026-08-07 · Owner: pierre
+
+Two open items lifted out of `install-path-gap.md` so they can be decided on their own. Both were
+proposed there while diagnosing an afternoon of `configd` updates that reported success and changed
+nothing, and both are about the same moment: the board is in a state you need to get out of, and the
+command you would reach for is a no-op.
+
+Neither is urgent any more, and that is deliberately the first thing said here. The restart set is
+now derived from the release (`engine.rs::units_to_restart`) and `updaterd` and `btd` restart
+themselves a few seconds after the update replies (`RESTART_AFTER_REPLYING`), so the situation these
+dig you out of is much harder to reach than it was when they were written. They are recorded because
+they are still *right*, not because anything is on fire.
+
+## 1. `already_current` compares the installed release, never the running one
+
+`Engine::apply` returns `ApplyResult::AlreadyCurrent` when the manifest's version equals the
+installed version (`engine.rs:366`, and again at `:833` for `select`). That is the correct answer to
+the question it asks. It is the wrong question in exactly one case: the release is installed and the
+process serving is from a different one.
+
+That case cost two hours. The update reported success, the daemon kept answering on old code, and
+`robotctl update apply` then said `already current` and did nothing — so the obvious recovery command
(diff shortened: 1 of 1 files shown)


pierre-rouanet:
Closing: the decision this asked for has been made, and shipped.

The document's central question was whether a running/installed mismatch should be
**reported and refused** or **repaired**, with a recommendation to repair. `updater/src/reconcile.rs`
(#64, #65, #73) is that answer, implemented — just at a different moment than `apply`:

- every daemon publishes an identity file at startup (`duck_ipc_proto::publish_identity`), `btd`
  and `padd` included;
- at each `updaterd` start, `Engine::reconcile_running_units` compares each shipped unit's running
  release against its component's active one and restarts what is stale;
- `updaterd` restarts itself five seconds after every applied update (`RESTART_AFTER_REPLYING`), so
  that check runs seconds after every `apply`. The state this document was written about now
  self-heals, with `apply` unchanged.

Two of its premises are no longer true. It says a `btd` running revision "cannot be read at all"
because `btd` serves no socket — the answer is a file now, so it is readable, and `robotctl health`
warns on it (`unit_warnings`). And it reasons about `install --force`'s guard as the precedent for
item 2; #68 changed what that guard treats as "answering", though not the conclusion drawn here —
`apply --force` still must not copy it by symmetry.

What survives is smaller than a decision document, and narrower than either item as written:

- `Engine::apply` still returns `AlreadyCurrent` on the installed version alone (`engine.rs:399`);
- `reconcile` deliberately will not repair one skew — `updaterd`'s own, because a self-restart loop
  in the process that owns recovery has no way out (`Verdict::ReportedOnly`). That is the one cell
  where an inert `apply` still costs something: a stale `updaterd`, an operator running `apply`,
  `already_current`, no restart scheduled, nothing to repair it but a hand-run
  `systemctl restart updaterd`.

That residue is one narrow change to `apply`, not two open questions, and `docs/design/restart-order.md`
§5 and §7 already own the mechanism and its diagnosis — a page at the `docs/` root restating it would
duplicate them.

PR #45, opened 2026-08-07
Decision wanted: three places one board disagrees with itself about a contract

pierre-rouanet:
The `hello` handshake direction, an unreadable `safeToRestart`, and a downgrade that
orphans a unit. Same shape each time: two things on one board disagree about a
contract, and the code chooses between refusing and coping. The choice has been made
implicitly, in three different directions, by three pieces of code that never compared
notes.

Two are lifted out of `install-path-gap.md` (§2.2 and the `203/EXEC` consequence) so
they can be decided outside a post-mortem. The middle one is new, found while adding
`Health::Incompatible`:

**`SafeToRestart` guesses "safe" on a reply it cannot parse, and its comment says that
is exactly what must not happen.** The comment reads "treated as unreachable rather
than guessed at: guessing 'safe' could restart a walking robot" — but
`permits_restart` is `!matches!(self, No(_))`, so `Unreachable` *is* safe. Both halves
are individually defensible, which is why this is a decision and not a patch: absence
meaning safe is correct and documented for its own case, because a `robotd` that does
not answer has no control loop running. A robot that *answers* is a different
situation and may well be walking. One of the comment or the code is wrong today
whatever gets decided, which is why it is recommended first.

For the other two, the write-up is mostly about what the original proposals left out:

- accepting older clients quietly promises that `API_VERSION` bumps are **additive**,
  and nothing states that rule. Either it gets written down, with a different mechanism
  for a non-additive change, or the promise is one the protocol cannot keep;
- a preflight check on orphaned units has to choose where it reads the installed unit
  list from — `/etc/systemd/system` includes units no release shipped, while the
  previous release's `systemd/` misses the orphan by construction — and whether it
  refuses or warns, since a deliberate downgrade past a daemon boundary is how you get
  off a bad release that introduced one. It also cannot inherit `WouldDowngrade`'s
  `Latest`-only exemption: `Ref` is how the case was observed.

Each item records that its urgency dropped once `updaterd` began restarting itself, so
nobody reads these as outstanding incidents.

No code. Item 1 changes what the daemon accepts from every client, item 2 changes when
an update may proceed on a moving robot, and item 3 can make a legitimate downgrade
impossible.

**No code in this PR by design.** Item 2 is the one to look at first — it is a live contradiction between a comment and the code it describes, found while writing #42.

### DIFF
--- a/docs/contract-mismatch-policy.md
+++ b/docs/contract-mismatch-policy.md
@@ -1,0 +1,129 @@
+# What to do when the other side speaks a contract you do not
+
+Status: **decision wanted, nothing implemented** · Date: 2026-08-07 · Owner: pierre
+
+Three open questions, all the same shape: two things on one board disagree about a contract, and the
+code has to choose between refusing and coping. None is a bug report — each is a policy call that has
+been made implicitly, in three different directions, by three pieces of code that never compared
+notes.
+
+Lifted out of `install-path-gap.md` (§2.2 and the `203/EXEC` consequence) plus one found while
+fixing the health-reply parsing.
+
+## 1. `hello` refuses a client that is merely *older*
+
+`updater/src/ipc.rs:493` is an exact `!=` on `API_VERSION`. Any disagreement in either direction is
+`PROTOCOL_MISMATCH`.
+
+`robotctl` is a symlink into `current`, so it follows the installed release. `updaterd` did not, until
+it began restarting itself after an update — so a release that changed `API_VERSION` guaranteed the
+two would disagree, and the command that stopped working was `robotctl update apply`, which is
+exactly the one you would use to get out of it. Observed by ordinary use, not contrivance: install
+branch A, install branch B while waiting for CI, and every call fails. `API_VERSION` has since
+reached 4, and the failure was seen again at v4-against-v3.
+
+The handshake at least *named both versions*, which is more than the health-parse failure managed.
+
+**The proposal.** Refuse only when the client is **newer** than the daemon. A v4 daemon serves a v2
+client perfectly when v4 only *added* methods; refusing that direction costs the ability to recover
+and buys nothing. Client-newer-than-daemon stays a hard failure — there the client may ask for
+something that genuinely is not there.
+
+That change makes `API_VERSION` mean what it should: "the newest contract I understand", not "the only
+contract I will speak". Additive protocol growth stops being a breaking change for every client on the
+box.
+
+**What needs deciding.** The proposal rests on a premise nothing enforces: **that `API_VERSION` bumps
+are additive.** Today the constant is a single number with no stated rule, and this change would quietly
+promise backward compatibility on every past and future bump. Either the rule gets written down and a
+non-additive change gets a different mechanism, or accepting older clients is a promise the protocol
+cannot keep.
+
+Urgency dropped a lot when `updaterd` started restarting itself — the window is now seconds rather than
+until someone reboots. This stays proposed on design grounds.
+
+## 2. An unreadable `safeToRestart` is treated as safe
+
+Found while adding `Health::Incompatible`, and it is the same class in the neighbouring method.
+
+`SocketRobotClient::safe_to_restart` maps a reply it cannot parse to `SafeToRestart::Unreachable`
+(`updater/src/robot.rs`), and its comment gives the reason:
+
+> An answer we cannot parse is treated as unreachable rather than guessed at: guessing "safe" could
+> restart a walking robot.
+
+But `SafeToRestart::permits_restart` is `!matches!(self, No(_))` — so `Unreachable` *is* safe. The
+comment describes the opposite of what the code does. The code guesses "safe", which is the outcome
+the comment says must be avoided.
+
+Both halves are individually defensible, which is why this is a decision and not a patch:
+
+- `Unreachable` meaning safe is **correct for its own case** and documented as such: a `robotd` that
+  does not answer is a `robotd` whose control loop is not running, so nothing is moving, and that is
+  precisely when an update is the fix. Making absence an error would block recovery on the robots that
+  need it most.
+- A robot that *answers* is a different situation. It may well be walking. Reading its "no, I am
+  mid-task" as "sure, go ahead" because a field was renamed is the failure the comment was written to
+  prevent.
+
+**The proposal.** Split them, the way `Health::Incompatible` splits an unreadable health reply from an
+absent daemon: a parse failure on a reply that *arrived* becomes its own variant, and it does not
+permit a restart.
+
+**What needs deciding.** Whether an unreadable answer should block an update, given that the whole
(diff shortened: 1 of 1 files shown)


pierre-rouanet:
**Item 2 is implemented in #68**, with the decision the doc left open resolved as: block, and no new bypass, because `systemctl stop robotd` already is one and is a better one. It converts an unreadable answer into an absent one — honest rather than a workaround, since a stopped robot cannot be moving and cannot be misread about it either — so it needs no flag, no config and no `API_VERSION` bump. The refusal text names it, which matters because this refusal happens on a robot that looks perfectly healthy.

Worth folding back into this doc when it next moves, along with two things found while implementing:

**Item 2 is narrower than it reads.** `SafeToRestartResult` has exactly one required field (`safe`; `reason` is already defaulted), so no robotd in existence trips the new variant — only a real shape disagreement does. The class that motivated it is `#[serde(default)]` on *nested* reply structs, which is `install-path-gap.md` §1 and still open. That does not weaken the fix, but it does mean this closes a latent hole rather than a live one.

**Item 3's mechanism does not fit where it is stated.** "Preflight refuses a target that lacks a binary some installed unit execs" cannot run in preflight: both passes run before the artifact is downloaded, so the candidate's file list does not exist yet — the first has no manifest at all and the second has only the manifest. The check has to run after extraction and before the swap, which is still "no side effects, no rollback churn" (staging is disposable, nothing is armed, `current` has not moved) but costs a download to find out.

That also answers one of the three open sub-questions in item 3, in passing. On **where to read the unit list from**: `/etc/systemd/system/*.service` is the right source *filtered to units whose `ExecStart` points into the component's `install_dir`*. That keeps the orphan — which outlived the release that installed it, so the previous release's `systemd/` cannot name it — while excluding units no release of ours ever shipped, which was the objection to reading the live directory. The remaining two are genuinely open: refuse-with-override versus warn, and whether it inherits `WouldDowngrade`'s `Latest`-only exemption (it must not — `Ref` is how the case was observed).

**Item 1** is untouched and should stay that way until the additive-bump rule is written down. Accepting older clients silently promises backward compatibility on every past and future bump, and there is one user and one robot today — so the cheaper answer may be to keep the exact `!=`, bump `API_VERSION` freely, and name the install consequence, rather than to make a promise the protocol has no mechanism to keep.

pierre-rouanet:
**Item 3 is implemented in #75.** Of its three open sub-questions, the one this PR left genuinely open — refuse or warn, and where the override lives — is resolved the way #68 resolved item 2: refuse, with no new override, because removing the unit already is one and is a better one. `systemctl disable --now configd.service && rm /etc/systemd/system/configd.service` makes the situation true rather than overriding a check that says it is not — a board below the release that introduced a daemon should not be running that daemon — and nothing is lost by it, since the next update that ships the unit reinstalls it. The refusal text names it.

The other two are settled as this thread already had them: the unit list comes from `/etc/systemd/system/*.service` filtered to units whose `Exec*=` points into the component's `current` symlink, and no target is exempt, unlike `WouldDowngrade`'s `Latest`-only guard.

Two corrections to fold back into the doc when it next moves, both found while implementing:

**The proposal's mechanism does not fit preflight**, as noted above — but the placement it moved to has a second constraint the doc does not anticipate. The check has to run *before the dry run returns*, not merely before the swap: `--dry-run` is how "will this downgrade work?" gets asked, and a check sitting at the swap would answer `DryRunPassed` to precisely that question. The cost is that a candidate cannot remedy its own orphan in its `preinstall` hook, which runs later. That is the right trade today — in the case that motivated this, the old release cannot know about the newer daemon, so self-remedy is impossible by construction — and it would want revisiting the first time a release deliberately drops a daemon.

**It also has to stay out of the recovery paths**, which the doc does not say. Rollback, reset-to-golden and `select` move backwards on purpose and are how a board gets off a bad release, so a check that can refuse must not sit there (`architecture.md` §1.1). Rolling back onto an orphaned unit stays `hooks/postinstall`'s documented, self-correcting behaviour.

One thing worth recording about scope: the check is not downgrade-specific. It fires on any candidate lacking a binary an installed unit execs, which includes an *upgrade* that drops a daemon while its unit stays behind — the same `203/EXEC`, and no reason to name it differently.

**Item 1 stays as this thread left it**: keep the exact `!=`, and spend the change on the message instead, which today names both versions and no remedy. Accepting older clients would promise additive-only bumps on every past and future bump with nothing to enforce it; naming the remedy costs a few lines and promises nothing.

pierre-rouanet:
**Item 1 is in #77**, resolved as this thread argued rather than as the doc proposed: the exact `!=` stays, and the change goes into the message.

The premise decided it. Bumps are not additive — v5 was, v4 was not, and `API_VERSION` does not distinguish them — so accepting older clients would promise backward compatibility on every past and future bump with nothing to keep it and no way to make a non-additive change afterwards. The constant now states that outright, along with what a bump costs: every client on a board must come from the same release as `updaterd`.

What was missing was never leniency. The refusal named two version numbers on a board where the two halves are a symlink and a running process, and said nothing about reaching either. It now branches on direction, because the direction *is* the diagnosis — client newer means the seconds after an update and the answer is retry; client older cannot happen through `/usr/local/bin/robotctl`, so the answer is to use that one. `robotctl` also stops appending "install matching versions", which was true and not actionable, and would now contradict a daemon that knows which side is behind.

With #75 that is all three items. This doc's own recommendation held for two of the three — item 2 first because it was unsafe rather than unhelpful, item 3 next — and item 1 inverted: it was the most defensible change and the one whose premise did not survive being written down, which is the argument for having written it down.

Worth folding into the doc before it lands, or closing it as the record of the decisions with #68, #75 and #77 as what came of them. I'd close it: all three are now stated where the code is, and a decision doc that outlives its decisions is the thing that gets read as still-open.

pierre-rouanet:
Closing. All three items are decided and shipped, and each decision now lives next to the code it governs rather than in a doc that has to be found:

- **Item 1** — `duck-ipc-proto`'s `API_VERSION` doc comment states the rule (a bump promises nothing in either direction, and every client on a board must come from the same release), and `updater/src/ipc.rs` names the remedy per direction. #77.
- **Item 2** — `SafeToRestart::Incompatible`, which does not permit a restart, with `systemctl stop robotd` as the escape rather than a new flag. #68.
- **Item 3** — `updater/src/orphan.rs`, refusing a candidate that lacks a binary an installed unit execs, with removing the unit as the escape. #75, and `updater-design.md` §7.3.

`install-path-gap.md`'s version-skew section is updated to match in #79, including the count of what is still open there.

Not merging this doc: two of its three recommendations were followed and item 1's was not, so as a page it now argues for a change that was decided against. The reasoning that survived is in the three PRs above and in the code. What this PR was for — getting the decisions made outside a post-mortem, one at a time — worked, and item 1 is the case for it: the proposal was the most defensible of the three, and writing its premise down is what showed the protocol could not keep it.

PR #52, opened 2026-08-10
docs: a design for seeing what the services say to each other

pierre-rouanet:
Design only — nothing implemented. `docs/design/monitor-design.md`, plus a pointer from `architecture.md` §8 and a row in the docs index.

## The shape

"Monitor the message exchange" is two requests: **the trace** (a developer at the bench, asking where an operation stopped) and **the meter** (support on a live robot, asking whether IPC is working and who is using it). One event stream, folded two ways — the same relationship `robotctl monitor` already has with `robot.state`. §9.1 adds a third fold: the last of both, on disk, for whoever arrives after the service died.

## The decisions

**The tap goes at the dispatch boundary, not on the socket.** An interposing proxy is the cheap design and disqualifies itself twice: `may_mutate` reads `SO_PEERCRED`, which behind a proxy reports the proxy's uid for every caller — so it either breaks mutating calls or grants any peer permission to change the robot — and it can only see one of the four transports §4.1 commits us to. Bytes on a wire also cannot say a call was refused by policy, or that a frame was dropped because a subscriber lagged.

**The shared serve loop comes first**, before any monitoring is written. A tap each new service must remember to call is exactly the `_ => None` wildcard `route.rs` refuses on the grounds that it is safe in the moment and wrong over time. The three loops differ in four ways that are requirements rather than drift — a peer policy in two of three, `updaterd`'s 1 MiB line cap against 64 KiB, a push side in two of three, two connection shapes — so this is a real abstraction over dispatcher, policy, cap and push stream. Underestimating it is the main risk in the plan, and three implementations is the cheapest it ever gets.

**Params are opt-in and redacted through the typed `Call`.** The tap sees the line before the hand-written `Debug` impls that keep the wifi PSK and the pairing PIN out of the journal. At dispatch the redaction is the natural path; at the socket it is a thing to remember. The on-disk record never carries params at all.

**The correlation id echoes on responses and notifications**, not only requests, so a captured line is self-describing — a log excerpt pasted into an issue reconstructs into a tree without the process memory that held the mapping. Omitted when absent; about a kilobyte a second on a 50 Hz stream. Needs a `PROTOCOL_VERSION` bump, so board and laptop move together.

**Every error response reaches `info`**, not only mutating ones, since read-only is most of the surface and its failures are invisible today. That hands a retrying client the ability to evict a journal, so the loop collapses identical consecutive errors and demotes a connection past an error budget — §8.1's retention argument applies here harder than it did there.

**The meter gets a crash record**: last snapshot plus a ring of the last N events, flushed to a fixed-size file under `/var/lib`, outside release dirs so it survives update and rollback. The ring is the part worth having — the trace cannot be left on permanently, and the events immediately before a service died are exactly what nobody can go back and collect.

## Surfaced by the refactor, deliberately not settled here

`robotd` passes no peer policy, and `Call::is_mutating()` does not cover `robot.enable` — so the call that starts a policy running on a walking robot is classified alongside `hello`. Defensible, since safety is authoritative regardless of what a client asks, but it has never been written down as a decision and it should not be answered by accident in a diff about serve loops. Noted with the consequence if revisited: `padd` sends `robot.enable`, so classifying intents as mutating denies the gamepad unless its user is allowlisted.

Other open questions: what `mediad` changes as the first service that is both client and gateway; how the trace ring is sized, given `padd` alone produces 50 events a second; and whether the shared loop is its own crate or a `duck-ipc-proto` module, which is really a question about keeping the types free of `tokio`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,6 +21,7 @@ the bug.
 | [`robotd-design.md`](design/robotd-design.md) | The control loop: model, bus, sensing, observations, policy, safety. |
 | [`updater-design.md`](design/updater-design.md) | The update engine: verification, atomic swap, health gate, rollback, release format. |
 | [`app-path-design.md`](design/app-path-design.md) | `btd` and `configd` — how a phone configures a robot over BLE. |
+| [`monitor-design.md`](design/monitor-design.md) | Seeing the messages the services exchange: the tap, correlation, the trace and the meter. |
 
 ## `project/` — you are running the project
 
--- a/docs/design/architecture.md
+++ b/docs/design/architecture.md
@@ -368,6 +368,10 @@ Deployment specifics — the journald drop-in, install steps, verification comma
 [`../deploy/README.md`](../../deploy/README.md). This section is the contract every service
 must satisfy.
 
+What a service logs about *itself* is here. What it logs about the messages it exchanges with
+the others — the tap, correlation across services, the trace and the meter — is
+[`monitor-design.md`](monitor-design.md).
+
 ### 8.1 Every service logs to stderr
 
 `tracing` → stderr → journald, level via `RUST_LOG` (`info` in the shipped units). No
--- a/docs/design/monitor-design.md
+++ b/docs/design/monitor-design.md
@@ -26,6 +26,9 @@ They are not two systems. They are **one event stream folded two ways**: the tra
 each event, the meter accumulates them. Building them separately is how a robot ends up with
 a log saying one thing and a counter saying another, and no way to tell which lied.
 
+There is a third fold, and §9 argues it is the most valuable one on a robot nobody can attach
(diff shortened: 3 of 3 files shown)


pierre-rouanet:
Closing my own draft.

The design still reads correctly to me — the tap belongs at the dispatch boundary and not
on the socket, for the `SO_PEERCRED` reason and because bytes on a wire cannot say a call
was refused by policy. But it is explicit that the shared serve loop comes first, and that
refactor is not on the table: three dispatch loops that differ in four ways is a real
abstraction, and nothing since August has made it more urgent than the work that has been
going in instead.

So this is a design doc for something nobody is about to build, sitting in the review
queue and conflicting with `main`. Better closed than left as a fifth stale entry.

Reopen when the serve-loop consolidation is actually being done; the doc is the right
place to start from at that point.

PR #53, opened 2026-08-10
Start brings the robot up, instead of a manual init nobody documented

pierre-rouanet:
Pressing Start on a fresh robot did nothing visible: the policy ran, the loop wrote positions, and the servos ignored them because they had no torque. Torque came from `robotd init` — a separate subcommand that opens the motor bus itself, so it needs the daemon stopped, and which appears in no documentation. The step existed, worked, and was invisible.

`robot.enable` now advances a bring-up state in the control loop:

```
Limp --enable--> Homing (torque on, 2s ramp to home) --> Ready --> policy drives
```

## The invariant that rule was protecting is unchanged

`duck-control` said the control loop "deliberately never touches torque so that a restart mid-update leaves a standing robot standing". That property is worth keeping and it does not need the broader rule: **nothing here happens because a process started.** Only an explicit `robot.enable` advances the state, which is a human pressing a button.

`a_restart_asks_for_no_torque` pins it on the *absence* of any write rather than a write of `false`, so an update restarting `robotd` mid-stand leaves the servos exactly as they were.

## Three gates, each for its own reason

- **A loaded policy.** `enable` means "enable the policy". Powering the joints to run one that is disabled or would not load would stand a robot up on a broken release and then hold it, which reads as working.
- **Not fallen.** `Safety::apply` commands a fallen robot at limp gain and holds it, so a ramp there would be writing a stand-up that cannot happen. `robot.enable` already refuses in that state; `robotd init` with the daemon stopped is still how you pick a robot up. (This was the explicit choice — the alternative was a sanctioned bypass of the fall clamp, which is a bigger change to the safety authority and deserves its own thinking.)
- **A fresh sample.** The ramp starts from where the joints are. Starting from a position nobody read is the lurch the ramp exists to avoid.

Torque is *not* dropped when the policy is disabled again — the robot holds its pose, which is what "a standing robot stays standing" means on this side too.

## Through the safety layer, not around it

`set_torque` joins the `RobotIo` trait and `Safety` gains a passthrough, because `Safety` owns the only handle — that is what keeps "nothing commands a motor except through here" a fact the borrow checker enforces rather than a convention.

It bypasses no clamp. Torque decides whether the motors hold what `apply` writes; every limit, the fall gate and the limp gain still decide *what* gets written. Worth contrasting with the status quo: `robotd init` bypasses the safety layer completely, because it opens the bus itself.

## Verification

`cargo test --workspace`: 471 passed, 0 failed. Clippy and fmt clean.

Three new tests, covering what CI can reach: the ramp arithmetic (starts where the robot is, halfway is halfway, and it terminates — a ramp that never finished would lock the policy out permanently), no torque at startup, and no torque when the policy is absent.

**Not verified on hardware.** The homing path needs a loaded policy and therefore a board, same as the policy path it gates. What wants checking there: that the two-second ramp from a resting pose to home looks right, and that Start-then-Start-again leaves the robot holding rather than dropping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ The controls:
 
 | | |
 |---|---|
-| **Start** | enable / disable the policy — nothing moves until this is on |
+| **Start** | enable / disable the policy — nothing moves until this is on. On a limp robot this is also what powers the joints: torque on, two seconds to the home pose, then it drives |
 | **Y** / triangle | switch between driving the **body** and posing the **head** |
 | **B** / circle | stop |
 | left stick | body: forward/back and strafe · head: neck pitch and roll |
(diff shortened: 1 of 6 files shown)


pierre-rouanet:
Folded into #49 — both changes are the same thing from the operator's side (what has to happen between a robot on the bench and a robot walking), so they are one thing to test. #49 now also carries `robot.init` / `robot.relax`, which is what makes powering the joints a command rather than a subcommand that fights the daemon for the motor bus.

PR #64, opened 2026-08-11
health: which release each daemon is actually running

pierre-rouanet:
Two gaps, one mechanism.

**`robotctl health` reported three daemons and stayed silent about two.** `btd` and `padd` serve no socket, so there was nothing to ask them — the code said so and sent people to `systemctl status btd`.

**After an update, "is this the new binary or the old one" had no answer for those two.** Every crate in the workspace shares one version, so on the dev channel the old build and the new one both read `0.4.0`. The only thing that tells them apart is the release directory the binary came from.

## How it is answered

Not by asking the process. Whatever mechanism it would answer through, **the process that needs interrogating is the old one**, which predates the mechanism. So it is read from outside: systemd knows `MainPID`, `/proc/<pid>/exe` resolves the `current` symlink to the real path, and a release installs to `…/releases/<version>/bin/`.

`configd` reads it, for two reasons: reading another user's `/proc/<pid>/exe` needs privilege `robotctl` does not have (`btd` runs as `btd`), and `configd` already asked systemd this exact question for `padd`. Its module doc had the argument already — a connected pad and a dead `padd` is the failure that looks like working hardware — and a dead `btd` is a robot no phone can see, which is the same silence. `driver.rs` becomes `units.rs` and answers for every unit a release manages.

```
units
  updaterd  active · 0.4.0
  robotd    active · 0.4.0
  configd   active · 0.4.0
  btd       active · 0.4.0-dev.271.7610e6e · binary deleted
  padd      stopped
```

A new `system.services` call carries it, routed over BLE too — an app that can trigger an update should be able to show whether it took.

## What warns and what does not

Only a version disagreement, and only for the daemons no socket covers. `updaterd`, `robotd` and `configd` report their own build when asked, which is better evidence than a path; warning from both sources would print one problem twice in two wordings. A stopped unit prints `stopped` and nothing else — a robot whose owner has no gamepad and disabled `padd` should not be scolded on every health check. Nothing here touches the exit code, so `install.sh`'s gate is unchanged.

## Two stale pieces of advice, corrected

Both described a world where the engine skipped `updaterd` and `btd` and left them for the next boot:

- `updater-design.md` still said those two "pick up the new binary at the next boot or an explicit later restart".
- `robotctl`'s own warning told you `updaterd` "keeps running the old binary until the next reboot".

`RESTART_AFTER_REPLYING` has scheduled both restarts 5 s after the reply for a while, so that advice sends someone to reboot a robot that had already fixed itself. Confirmed on the board while writing this: `restart scheduled unit="btd" delay="5s"`, and every daemon came up on the installed release within eleven seconds. The test pinning that wording now says why it is pinned.

## Checked

- Workspace suite green; six new tests covering the path parse (dev suffixes, deleted binaries, hand-built binaries outside the layout) and the warning rules.
- `cargo clippy --all-targets` clean on host and `aarch64-unknown-linux-gnu` for every crate touched.
- `DriverState` → `UnitState`: a Rust-only rename, since those serialise as their own names.
- **Not yet run on the board** — the `/proc` read and the systemd `MainPID` query need an install to exercise. `robotctl health` after one is the check.


pierre-rouanet:
Reworked, per the call to stop designing around compatibility. The `/proc/<pid>/exe` mechanism is gone: **each daemon now publishes its own identity at startup** to `/run/<service>/identity.json`, from the shared macro that already logs those exact facts.

An old daemon is allowed to simply not answer — it reports `build unknown (old)`, and stops doing so the first time it restarts.

What that changed, in order of how much it matters:

- **The revision is now available.** A path names a release directory; a process knows its own git SHA. On the dev channel two releases share a crate version, so the SHA is the whole difference — previously reachable only by inference from a directory name that happens to embed it. The unit lines read `active · 0.4.0 (rev 7610e6e)`.
- **No privilege anywhere.** Reading another user's `/proc/<pid>/exe` needs root; reading your own needs nothing. That single fact was the entire justification for `configd` having to be the one to answer, and both the code and the paragraph defending it are gone.
- **Less machinery.** No `MainPID` query, no `/proc` read, no PID plumbing, no `deleted` field. `configd` keeps only the question systemd alone can answer — is the unit running — which it already asked for `pad.status`.

Mechanics: `RuntimeDirectory=<service>` in each of the five units. It is the only thing that both creates a directory owned by the unit's `User=` *and* survives `ProtectSystem=strict`, which otherwise leaves `btd` and `padd` with a read-only filesystem — and systemd removes it with the unit, so a stopped daemon cannot leave an identity behind claiming to be running. No tmpfiles.d, no packaging changes.

`DUCK_RUNTIME_DIR` overrides the root. Nothing on a robot sets it; it is what makes publish-and-read testable, since a test cannot write to `/run`, and it is also how a daemon run by hand on a laptop publishes at all. Verified end to end locally: `padd` writes `{"service":"padd","version":"0.4.0","exe":"…/target/debug/padd","pid":36725}`.

Also here: the dev cheat sheet's manual recipe — `readlink /proc/$(pgrep -x configd)/exe` against the `current` symlink — becomes `robotctl health`, which was the point of the exercise.

#65 is rebased on this and now reads the same files instead of shelling out to `systemctl show` and readlinking.

PR #65, opened 2026-08-11
updaterd: check the restarts happened, and fix what did not

pierre-rouanet:
Stacked on #64 — it reuses the `/proc/<pid>/exe` mechanism from there. Review that one first.

**Scheduling a restart is not evidence one happened, and scheduling is all that was ever checked.** `systemd-run` returning success means a transient timer was created — not that the restart ran, not that the new binary started. Those failures are logged and swallowed on purpose, and rightly: an update that worked must not report failure over a restart it could not arrange.

That leaves exactly one way for a robot to end up running a release it did not install, and it is silent. It also lands on the two units with nothing watching them, because they are the two an update cannot restart while it is running: `updaterd` and `btd`.

## What it does

Every `updaterd` start compares each unit's running binary against the release its component has active, and restarts anything stale.

Startup is the first moment that answer can be trusted: `updaterd` cannot watch its own replacement land, so the check belongs to the successor. It catches more than a missed timer — a restart that failed, a binary that would not start, a unit stopped mid-update, a rollback that left one daemon behind. At boot it is a no-op, since everything starts from the same symlink, and it costs one `systemctl show` and one `readlink` per unit.

## Two things it deliberately does not do

- **It never restarts `updaterd` itself.** A successor that disagreed about which release is active would restart, disagree again, and loop — in the one process that owns recovery, so nothing would be left to break the cycle. Reported, not acted on, which is safe: it has just started, so anything stale about it was decided before this code ran.
- **It leaves a stopped unit stopped.** Someone stopped it. Starting it here would override that on every `updaterd` start — most likely against whoever has no gamepad and disabled `padd`.

## Restarting rather than only reporting

You asked for the check; this also fixes what it finds. Reporting alone leaves a robot running the wrong code with a warning in a journal nobody reads, which is the situation this exists to end. The acting part is one call site (`reconcile::check`) if you would rather it only reported.

## Notes

- `release_from_path` moves to `duck-ipc-proto` so `configd` and `updaterd` cannot drift on what a release path means. It is pure and needs only `semver`, which that crate already has.
- The decision is a pure function with the syscalls kept out of it, and that is the point: a unit running a release that is not the active one is a state only a broken update produces, so it cannot be arranged on a board and has to be testable without one. Five tests cover stale, dev-vs-stable suffixes, self, stopped, and current.
- Workspace suite green; clippy clean on host and `aarch64-unknown-linux-gnu`.
- **Not exercised on hardware** — it needs an install, and the interesting path needs an install whose restart failed.

### DIFF
--- a/updater/src/engine.rs
+++ b/updater/src/engine.rs
@@ -1082,6 +1082,46 @@ impl Engine {
     /// [`MAX_BOOT_ATTEMPTS`] boots, and delete staging leftovers. This is the path
     /// that catches a release which doesn't start at all — the in-process health
     /// gate can't, because it died with it.
+    /// Check that the restarts an update scheduled actually happened, and fix what did not.
+    ///
+    /// Runs at startup, beside [`Self::recover_on_start`], because that is the first moment the
+    /// answer exists: `updaterd` cannot watch its own replacement land, so the check belongs to the
+    /// successor. See [`crate::reconcile`] for why scheduling a restart is not evidence one
+    /// happened.
+    ///
+    /// Per component, since each has its own active release and ships its own units. Failures are
+    /// logged rather than returned, for the same reason the scheduling is: this must never be a
+    /// reason to refuse to serve.
+    pub async fn reconcile_running_units(&self) -> Vec<crate::reconcile::Finding> {
+        let mut findings = Vec::new();
+
+        for name in self.config.components.keys() {
+            let Ok(store) = self.store(name) else {
+                continue;
+            };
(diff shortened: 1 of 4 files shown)


pierre-rouanet:
Rebased onto the reworked #64. `reconcile` now reads the identity each daemon publishes instead of `systemctl show -p MainPID` plus a `/proc` readlink — so the check is one file read per unit, and it gained a third refusal: a daemon that published nothing is left alone, because that is a build too old to say what it is rather than a stale one. Restarting a robot's daemons for being old is a decision nobody asked for.

PR #70, opened 2026-08-11
A test plan that costs no CI time

pierre-rouanet:
Rewrites `install-path-gap.md`'s "What would close it". Documentation only.

The A–D ranking it replaces was written before CI became the slow part of iterating, and it ranked purely on coverage — "real work: cgroup and privilege setup in CI, and slow" was the whole of its cost analysis. Re-ranked with the wait as a first-class constraint, and the shape of the answer changes.

## The constraint, made explicit

The four jobs run in parallel, so the wait for green is the **slowest job, not the total**. That decides where a new test belongs more than its coverage does:

- an in-process test in `cargo test` costs nothing anybody notices;
- anything landing in `board` (zigbuild from source, QEMU) or `coverage` (a full instrumented build) is paid on every push.

Three rules follow, and they are the part worth agreeing to rather than the item list: the default `cargo test --workspace` takes only in-process tests; anything that unpacks an artifact or drives a service runs **on demand**, never on the pull-request path; and **no new CI job**.

That last rule changes the biggest item. "Install the artifact in a container" was a day of work and a new job; it becomes a step hanging off the `xtask package` that `check` already runs, unpacking the tarball that step already built into a temp root with a stub `systemctl`. Seconds of shell, no second build.

## And something to remove

`coverage` runs the whole instrumented suite **twice** on a pull request — head, then base — to produce a delta comment. The absolute `--fail-under-lines` gate is what catches a regression; the delta is a nicety on the critical path of every push. Dropping the second run roughly halves the job, which buys more iteration speed than everything in this plan adds.

Not done here — it is a change to `ci.yml` and belongs in its own PR if you agree with it.

## Two items that did not exist when the first draft was written

Both are in-process, both cover mechanisms whose *acting* half has no test at all:

- **`schedule_deferred_restarts` hardcodes `Command::new("systemd-run")`**, while `SYSTEMCTL` two functions away is a `const` precisely so a test can substitute a stub. Nothing can observe that call today — `--on-active=5s` could be wrong and every test would pass.
- **`reconcile::check` (#65) is never called by a test**, only its pure `verdict_for`. It already takes `systemctl` as a parameter and reads identities through `DUCK_RUNTIME_DIR`, a seam whose own comment says it exists so this is testable.

## Housekeeping

The old A–D ranking is **removed** rather than left beneath the new plan — two plans for one question is exactly the drift the docs index now has a rule about (#66). Its still-true facts are folded into the item that replaces them: the three source-reading tests that compare two files to each other, and `board-test.sh`'s comment about `install.sh` above a line that runs `updaterd install --from` instead.

The three references to the letters elsewhere now name sections instead. One more lives in #36's document (`needs install-path-gap.md option C`) — worth updating there when that PR next moves; its point stands, and it is now item 4.

Depends on nothing. Reads best after #66, which is where the "one page owns a mechanism" rule lands.

### DIFF
--- a/docs/design/app-path-design.md
+++ b/docs/design/app-path-design.md
@@ -456,8 +456,9 @@ drifted from it**:
 So the trait is not merely a testing seam; it is the only written form of the contract. The problem is
 the direction of the check — the suite verifies the fake against the contract, and nothing verifies NM
 against either. Both implement `Net`, so the same assertions *could* run against both; what stops it
-is that the NM side needs a real NetworkManager and a real radio, which means a board
-(`install-path-gap.md`, option D) rather than CI.
+is that the NM side needs a real NetworkManager and a real radio, which means a board rather than CI —
+see `install-path-gap.md` §"What would close it", where a board is on-demand by design and deliberately
+not a CI runner.
 
 Until that exists, the honest summary is: `configd`'s wifi behaviour is tested, and the code that runs
 on the robot is not. Every bug above was found by hand, on hardware, in the space of one session.
--- a/docs/project/install-path-gap.md
+++ b/docs/project/install-path-gap.md
@@ -152,53 +152,108 @@ two files agree with each other, not that the thing they produce is correct.
 
 ## What would close it
 
-Roughly in order of cost.
+Revised 2026-08-11, after the restart mechanisms below landed and with one constraint that was not
+stated the first time: **CI is already the slowest part of iterating, so the plan is judged on what it
+adds to the wait, not only on what it covers.** Everything here therefore names where it runs.
 
-**A. Assert the artifact's contents.** Run `xtask package` in a test, then inspect the tarball:
-every unit named by `install.sh` is present, and every unit's `ExecStart` binary is present. Cheap,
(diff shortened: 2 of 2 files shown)


pierre-rouanet:
Two corrections to the description above, both after the fact.

**The "one page owns a mechanism" rule is #71, not #66.** It was meant to be #66's third commit and landed on the branch a few minutes after that merged, so it was stranded. Same change, rebased.

**A second removal, added in `8b4c104`, and this pull request is the evidence for it.** `on: [push, pull_request]` has no path filter, so editing a markdown file cross-compiles for aarch64 under QEMU and builds the workspace twice under instrumentation. Three consecutive docs pull requests — #66, #70, #71 — each paid that in full while the plan arguing against slow CI was being written.

The fix is a `paths-ignore` for `docs/**` and `*.md`, and the doc names the trap alongside it: a skipped job never reports a status, so filtering a check that is *required* for merge leaves docs pull requests permanently pending. Worth checking the branch protection settings first rather than finding out on a branch nobody can merge.

Neither removal is done here — both are `ci.yml` changes and belong in their own pull request if you agree with them.

PR #83, opened 2026-08-11
Drive a real update against real systemd

pierre-rouanet:
Item 4, the last of the plan in `install-path-gap.md` — and it found a bug in the engine before it had asserted anything.

## What only this can see

Every other check in the tree drives the engine against a **stub `systemctl`**. They cover the decisions well: which units it would restart, in what order, what it does when one is missing. None of them can see whether a restart *happened*.

`scripts/systemd-test.sh` boots systemd as pid 1, mints three signed releases carrying real units and a real `updaterd`, installs one, applies the next through the running daemon, then applies one that ships a unit which cannot start:

```
    [ok] systemd is pid 1 (degraded)
    [ok] postinstall installed, enabled and started units on a board that had none
    [ok] 1.0.0 is live
    [ok] 1.1.0 is live
    [ok] on_apply restarted the daemon the release ships (pid 154 -> 297)
    [ok] the transient timer replaced updaterd after 4s (pid 173 -> 356)
    [ok] the restarted updaterd is running 1.1.0
    [ok] its startup reconciliation restarted nothing, as it should
    [ok] the update rolled back rather than reporting success
    [ok] rolled back to 1.1.0
    [ok] the reason names the unit, not 'unreachable'
    [ok] the daemon is running again after the rollback
```

Four of those are new information:

- **`on_apply` really restarts**, asserted on the unit's main PID changing rather than on a command having been issued.
- **The deferred transient timer really fires and really replaces `updaterd`**, and the successor runs the new release. A child process could not do this — it would sit in the cgroup being killed — which is the entire reason for `systemd-run` and was until now an untested claim in a comment.
- **`hooks/postinstall` really installs, enables and starts** a unit the board has never had. The plan noted no stub can show this.
- **A unit that installs cleanly and cannot start fails the update and names itself**: `restart failed: Job for broken.service failed…` rather than `not healthy within 30s: unreachable`. That is bug 1 of `install-path-gap.md`, and the only class the cheaper items cannot reach.

## The bug it found

The first commit is a fix, not a test. `self_test_updaterd` ran the release's new binary with **no `--config`**, and that flag has a default — so the probe validated `/etc/robot/updater.toml` however the running daemon was started. The rollback reason was:

```
the new updaterd failed its self-test: invalid config
path=/etc/robot/updater.toml error=No such file or directory
```

blaming the release for a file it has nothing to do with. Worse, it defeats the check's stated purpose: catching a new binary that rejects *the board's* `updater.toml`, the operator's file preserved across installs, which it cannot do while pointed elsewhere. `Config` now remembers where it was read from, and the probe is given it.

Harmless on a board today, since the unit passes the default path. Not harmless on a dev box using `--config`, and not harmless as a guarantee.

## Docker rather than `systemd-nspawn`

A deliberate change from what the plan proposed, argued in the script's header. The objection to privileged containers **in CI** stands, and this is not CI. What decided it is the plan's own reason for ranking real systemd last: *a check that can only run on a machine nobody develops on stops being read.* Docker Desktop runs a privileged container with systemd as pid 1 on the laptop this was written on; nspawn does not. The fidelity that matters — real units, real cgroups, real transient timers — is identical.

## Notes

- **Not in CI**, deliberately: it needs `--privileged` and the host's cgroup filesystem, and `board` is already the slowest job. Run it when `on_apply`, the deferred restarts, or the hook change.
- The fixture is a new example rather than a flag on `fake-release`, which deliberately mints releases with **no** units because it exists to drive the engine where systemd is absent.
- `fake-robotd` is a `sleep`: what is under test is whether the engine restarts what a release ships, and a sleep proves that as well as motor control while needing no hardware. It publishes no identity, so the reconciliation sees `Unknown` and leaves it alone — which is why that assertion is written as an absence.
- `-p updater` green (150 lib tests), clippy clean, and the harness lints under the `shellcheck` gate added in #78.
- Two injections remain, both cheap now the harness exists: `systemd-run` failing, and `enable --now` on a unit whose `User=` does not exist.

With this, every item of the plan is done.

### DIFF
--- a/docs/project/install-path-gap.md
+++ b/docs/project/install-path-gap.md
@@ -312,17 +312,41 @@ behaving under `ProtectSystem=strict`, real unit states reaching `robotctl healt
 reconciliation closing the loop for real. For the timing-dependent parts a board is *higher* fidelity
 than any container, and it is now cheaper than one.
 
-### 4. Real systemd, locally, for failure injection only
+### 4. Real systemd — done, as `scripts/systemd-test.sh`
 
-`systemd-nspawn` on a Linux box — **not** a privileged container in CI. After the three items above,
-what is left is a short list of things nobody should ask a board to do repeatedly: a unit that exists
-and will not start, `kill -9` between the swap and the commit, `systemd-run` itself failing,
-`enable --now` on a unit whose `User=` does not exist. **Would have caught bug 1**, which is the only
-one the items above miss — and it is also the only way to watch the postinstall hook do its real job
-of enabling and starting a unit, which no stub can show.
+**Done, and it paid for itself before asserting anything.** `scripts/systemd-test.sh` boots systemd as
+pid 1, mints three signed releases carrying real units and a real `updaterd`, installs one, applies
+the next through the running daemon, and then applies one that ships a unit which cannot start.
 
-Last for a reason beyond cost: it can only run on Linux, so never on the machine this is developed on,
-and a test that runs somewhere else stops being read.
+Four things it observes that nothing else in the tree can, because everything else has a stub
+`systemctl`:
+
+- **`on_apply` really restarts** what the release ships — asserted on the unit's main PID changing,
+  not on the command having been issued;
+- **the deferred transient timer really fires and really replaces `updaterd`**, in about four
+  seconds, and the successor is running the new release. A child process could not do this: it would
+  sit in the cgroup being killed, which is the reason for `systemd-run` and was until now an
+  untested claim;
+- **`hooks/postinstall` really installs, enables and starts** a unit the board has never had;
+- **a unit that installs cleanly and cannot start fails the update and names itself** —
+  `restart failed: Job for broken.service failed` — rather than reverting with
+  `not healthy within 30s: unreachable`. That is bug 1, and it is the one class the items above
+  cannot reach.
+
+**Docker rather than `systemd-nspawn`**, which is a change from what this section used to propose.
+The objection to privileged containers *in CI* stands and this is not CI. What decided it is the
+argument this section itself made for ranking real systemd last: a check that can only run on a
+machine nobody develops on stops being read. Docker Desktop runs a privileged container with systemd
+as pid 1 on the laptop this is developed on; nspawn does not.
+
+It found a bug on its first run, in the code rather than in itself. `self_test_updaterd` ran the new
+binary with no `--config`, so it validated `/etc/robot/updater.toml` however the running daemon was
+started — defeating the check's whole purpose, which is to catch a new binary that rejects *the
+board's* config file.
+
+Two injections named here are still worth adding now that the harness exists, and neither needs new
+machinery: `systemd-run` failing (the update must still succeed, and the successor must reconcile the
+stale unit), and `enable --now` on a unit whose `User=` does not exist.
 
 ### Not doing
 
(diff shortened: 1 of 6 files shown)


pierre-rouanet:
Both injections added, in `804560c`. Eighteen checks now, all passing:

```
    [ok] systemd is pid 1 (degraded)
    [ok] postinstall installed, enabled and started units on a board that had none
    [ok] 1.0.0 is live
    [ok] 1.1.0 is live
    [ok] on_apply restarted the daemon the release ships (pid 177 -> 352)
    [ok] the transient timer replaced updaterd after 4s (pid 197 -> 411)
    [ok] the restarted updaterd is running 1.1.0
    [ok] its startup reconciliation restarted nothing, as it should
    [ok] the update still succeeded — an unschedulable restart is not an update failure
    [ok] the journal says the restart could not be scheduled
    [ok] btd and updaterd are left on 1.1.0, as a missed timer leaves them
    [ok] its reconciliation restarted the stale btd onto 1.2.0 after 0s
    [ok] and did not restart itself, which in that process would be a loop with no way out
    [ok] accounts before units: the shipped user exists and its unit is running
    [ok] the update failed and named the unit rather than the account
    [ok] current went back to 1.3.0
    [ok] the rollback failed too, which is the known gap: the bad unit file outlived the release
    [ok] the daemons that could restart did, so only the broken unit is down
```

**`systemd-run` cannot be run** — shadowed in `/usr/local/sbin`, which precedes the real one in the PATH systemd gives a unit. The update still succeeds, the journal says why, and `btd` and `updaterd` are left on the old release exactly as a missed timer leaves them. Then the next `updaterd` start restarts the stale `btd` and does *not* restart itself. That is `reconcile.rs` closing its loop for the first time outside a unit test — and it needed a fixture addition: only a unit actually named `btd` is in `NEVER_RESTART`, so the stand-in is named that, and unlike `fake-robotd` it publishes an identity because that is what the reconciliation reads.

**A unit whose `User=` the release brings with it** — `systemd-sysusers` runs before the units, the account exists, the unit runs as it. That ordering is called load-bearing in the hook's own comment and had never been observed. The other half, the same unit with nothing creating the user, fails the update and names the unit rather than the account.

## A real gap it turned up

**A failed update can fail its own rollback.** Applying the missing-user release gives:

```
error: rollback failed after a failed update: restart failed:
Job for needs-a-user.service failed because the control process exited with error code.
```

`postinstall` overwrote 1.3.0's good unit file with 1.4.0's broken one and, by design, does not put it back — the hook argues a release which did not take leaves one service failing until the next one does, the same situation either way. **That holds while the failing unit is only failing.** It stops holding when the unit is in the restart set of the release being reverted *to*, because then the revert re-runs the same restart and inherits the same failure. `RollbackFailed` is the outcome the design calls the most serious one.

Reachable by an ordinary bad release, not only by the downgrade case `install-path-gap.md` already records: two consecutive releases ship the same unit name and the newer one is broken. The tree does revert — `current` goes back, asserted — so what is lost is the apply action, not the release.

Asserted as current behaviour with a comment saying so, so that changing it is deliberate. The options are not symmetric and I have not picked one: making `postinstall` reversible is the change the hook explicitly declined, while tolerating a failed restart *during a rollback* weakens the distinction `restart_one` draws on purpose. Recorded in the doc as needing a decision.

## Two of my own assertions were wrong

Both corrections are in the diff, and both are about the code being right:

- **`ReportedOnly` is not reachable by restarting `updaterd`.** A restart re-execs through `current`, so the successor is the active release by construction. What the step actually proves is the absence of a *second* restart, which in that process would be a loop.
- **A rollback is a successful call reporting an unsuccessful outcome.** `robotctl` exits 0 having told you what happened, so the check is the reported outcome and never the exit status.

PR #84, opened 2026-08-11
A revert that happened is not a failed rollback

pierre-rouanet:
Fixes the gap `scripts/systemd-test.sh` found, and carries the two injections that found it — **`804560c` did not make it into #83**, which merged one commit earlier, so the harness on `main` is currently missing them.

## The bug

`rollback_to` mapped every step to `Error::RollbackFailed`, the apply action included. So a failed update whose failure was a bad unit file reported:

```
rollback failed after a failed update: restart failed:
Job for needs-a-user.service failed because the control process exited with error code.
```

That asserts the one thing that was false — the robot had not been put back — and buries the one thing that was true, which is that a daemon is down. `current` had already gone back; the harness asserts it. `RollbackFailed` is the outcome the design calls the most serious one, so saying it wrongly costs more than the words do.

**The cause is not the rollback.** `hooks/postinstall` overwrites unit files and, by design, does not put them back — the hook argues that a release which did not take leaves one service failing until the next one does, the same situation either way. That reasoning holds and is untouched. What it did not anticipate is the unit being in the restart set of the release being reverted **to**: then the revert re-runs the same restart and inherits the same failure. An ordinary bad release reaches it — two consecutive releases ship the same unit name and the newer one is broken — not only the downgrade case `install-path-gap.md` already records.

## The fix

**Separate the swap from the apply action, because only one of them is the recovery.** Past `swap_to` and the trial being cleared, the robot *is* back on the release it came from. A unit that then fails to restart is a second fact, not a contradiction of the first.

- `Error::RollbackFailed` keeps its meaning: the robot could not be put back at all.
- A failed apply action during a revert goes into the reported outcome and is logged at `error`:

```
not healthy within 30s: …; the release was reverted but a unit did not restart (…), so
something on this robot is down
```

Why the update failed comes first, because that is what someone is looking for; the daemon that did not come back is a second thing to act on rather than a correction of the first.

## The two alternatives, since neither is obviously wrong

- **Make `postinstall` reversible** — record what was overwritten so a revert can restore it. This is the change the hook explicitly declined. It needs state outside the release, and it adds a failure mode to the recovery path, which is the one place that has to stay boring. It also would not help the general case: a unit can fail for reasons that have nothing to do with its file.
- **Tolerate a failed restart everywhere** — this would weaken the distinction `restart_one` draws on purpose. A unit that exists and will not start is *how a broken release is caught*, and that stays fatal on the way in. Only the way out changes.

## Verification

`fail_rollback_apply` is a new fault, distinct from `fail_rollback`, because the two outcomes are now different and both need pinning. `a_revert_whose_restart_fails_is_still_a_revert` asserts the outcome, both halves of the reason, and that the tree really is back — the claim the outcome makes.

And the harness that found it now asserts the new behaviour, against real systemd:

```
==> applying 1.4.0, whose unit names a user nothing creates
    [ok] the update failed and named the unit rather than the account
    [ok] current went back to 1.3.0
    [ok] reported as reverted and as having left something down, not as a failed rollback
    [ok] the daemons that could restart did, so only the broken unit is down
```

Full harness green (18 checks). Workspace suite green, clippy clean, fmt clean.

## Note on the injections commit

`cf9c1df` here is `804560c` from #83, cherry-picked because that PR merged without it. It is unchanged apart from the assertion this fix inverts — it pinned the old behaviour deliberately, so that changing it would be a visible decision. This is that decision.

### DIFF
--- a/docs/project/install-path-gap.md
+++ b/docs/project/install-path-gap.md
@@ -367,24 +367,38 @@ Both injections named here are in it too, and both landed assertions nothing els
   account exists and its unit runs. The other half — the same unit with nothing creating the user —
   fails the update and names the unit rather than the account.
 
-### The gap the harness found on the way
+### The gap the harness found on the way — fixed
 
-**A failed update can fail its own rollback**, and `RollbackFailed` is the outcome the design calls the
-most serious one. Reproduced by the missing-user release above, and asserted as current behaviour so
-that changing it is deliberate:
+**A failed update reported a failed rollback for a robot it had successfully put back.** Reproduced by
+the missing-user release above, and `RollbackFailed` is the outcome the design calls the most serious
+one — so it was worth more than a note.
 
-`hooks/postinstall` overwrites unit files and, by design, does not put them back on a rollback — the
-hook argues that a release which did not take leaves one service failing until the next one does, which
-is the same situation either way. That holds while the failing unit is only *failing*. It stops holding
-when the unit is in the restart set of the release being reverted **to**, because the revert re-runs
-the same restart and inherits the same failure.
+The cause is not the rollback. `hooks/postinstall` overwrites unit files and, by design, does not put
+them back: the hook argues that a release which did not take leaves one service failing until the next
+one does, which is the same situation either way. That reasoning holds and is unchanged. What it did
+not anticipate is the unit being in the restart set of the release being reverted **to** — then the
+revert re-runs the same restart and inherits the same failure. Reachable by an ordinary bad release
+rather than only by the downgrade case above: two consecutive releases ship the same unit name and the
+newer one is broken.
 
-Reachable by an ordinary bad release rather than only by the downgrade case this document already
-records: two consecutive releases ship the same unit name and the newer one is broken. The revert
-itself does happen — `current` goes back — so what is lost is the apply action, not the tree. Worth a
-decision rather than a patch, and the options are not symmetric: making `postinstall` reversible is the
-change the hook explicitly declined, while tolerating a failed restart *during a rollback* weakens the
-distinction `restart_one` draws on purpose.
+**The fix separates the swap from the apply action, because only one of them is the recovery.** Past
+`swap_to` and the trial being cleared, the robot *is* back on the release it came from; a unit that
+then fails to restart is a second fact, not a contradiction of the first. So `Error::RollbackFailed`
+keeps its meaning — the robot could not be put back at all — and a failed apply action during a revert
+is carried into the reported outcome and logged at `error`:
+
+```text
+not healthy within 30s: …; the release was reverted but a unit did not restart (…), so
+something on this robot is down
+```
+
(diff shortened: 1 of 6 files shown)


pierre-rouanet:
Third commit, `fb1621e`: **an update was restarting the recovery net's oneshot.**

```
units_to_restart = ["configd", "padd", "robot-boot-check", "robotd"]
```

`robot-boot-check.service` asks whether the release that booted came up and hands over to `robot-rescue` if not. So an update restarting it points a rollback check at daemons that are legitimately mid-restart — and `robot-rescue` can answer that by swapping to golden and rebooting. During an update.

`hooks/postinstall` already declines to `enable --now` it, and says why in as many words:

> `enable --now` on it would run a rollback check in the middle of the update that installed it, with daemons legitimately mid-restart

The engine then read every `*.service` and restarted it a moment later anyway. Two places, one rule, one unit, and they disagreed — which is the shape of most of what this week turned up.

**What kept it harmless is a guard written for a different caller.** The script declines when uptime is over ten minutes, and its comment names `postinstall` as the reason that guard exists. Inside the first ten minutes of a boot — a dev board, or `auto_apply` polling sixty seconds after `updaterd` starts — the guard does not apply and the check evaluates its members while `configd` has just restarted and `robotd` is about to. It needs `failed` or three restarts to act, so it would usually decline. "Usually" is not the property you want from the thing that reboots the robot.

Note the ordering detail: `robot-boot-check` sorts *before* `robotd`, so it runs mid-sequence rather than after.

## The fix

Keyed on `[Install]` rather than on a name in `NEVER_RESTART`. A unit without that section is triggered by something else — a timer, or another unit pulling it in — so its lifecycle is not the update's to drive. Stating it as the rule makes `postinstall` and the engine agree by construction instead of by both being remembered, and the next unit like this one needs nobody to notice.

An unreadable unit file counts as one to restart: more likely a permissions problem than a triggerless unit, and a restart failing loudly beats one skipped quietly.

## Demonstrated, not argued

The fixture ships a unit shaped like the real one — no `[Install]`, and it appends to `/run/recovery-check.ran` each time it runs — and the harness asserts it did not run:

```
    [ok] on_apply restarted the daemon the release ships (pid 180 -> 359)
    [ok] and left the recovery check alone, which only a timer may start
```

With the rule disabled, that assertion fails and prints the `ran` the unit left behind:

```
    [ok] on_apply restarted the daemon the release ships (pid 180 -> 358)
ran
    [FAIL] the update ran a unit only a timer may start
```

So it catches the real thing rather than passing by construction. 19 checks green with the fix; `-p updater` 152 lib tests, clippy and fmt clean.

## Docs

`install-path-gap.md` is **retitled and closed** — its body has said the gap is shut since the artifact-install test landed, while its title and revision note still said otherwise, which is the exact drift the ownership rule exists for. `restart-order.md` gains §1.1 for the rule and stops calling five units the whole set, now that a release ships a sixth that is not a daemon.

PR #96, opened 2026-08-17
dev-push: sideload from a directory updaterd can actually read

pierre-rouanet:
`scripts/dev-push.sh` copied the release to `/var/tmp/duck-sideload`. `updaterd.service` sets `PrivateTmp=yes`, which gives the unit its own `/tmp` **and** its own `/var/tmp` — so the files landed in the pair the shell sees, and the daemon read the pair the namespace gave it:

```
==> copying to pierre@192.168.50.63:/var/tmp/duck-sideload
==> applying on pierre@192.168.50.63
  Preflight
error: network error: no manifest for version 0.5.1-dev.local.1786973040.g7fc1444 in /var/tmp/duck-sideload
```

`ls` on that directory lists exactly that manifest, its `.minisig`, the artifact and its `.minisig`. There is nothing in the error to act on, and nothing in the repo's tests to catch it: they point `LocalDir` at a fixture root, where no namespace applies.

### What changed

**The push sideloads from the board user's home**, resolved on the board (`$HOME` on a laptop is the laptop's). It keeps the one property `/var/tmp` was picked for — writable by the ssh user, so no sudo to copy into it — and is outside the unit's namespace. `updaterd` sets no `ProtectHome=`.

**Preflight gains `Check::SideloadDir`.** A `--from` directory this process cannot see now fails before any manifest lookup, and under `/tmp` or `/var/tmp` the message names the cause:

```
error: preflight check failed: SideloadDir: /var/tmp/duck-sideload is not there for updaterd,
whichever shell created it: this unit runs with PrivateTmp=yes, so it has a /tmp and a /var/tmp
of its own and neither is the one you copied into. The release is fine — put it anywhere outside
those two (a home directory, or /var/lib) and install from there.
```

A missing directory anywhere else stays a plain "does not exist" — mount namespaces are the wrong thing to send someone after when they typo'd a path. A board that does not privatise `/var/tmp` passes the check without noticing it exists, and `updaterd install --from` (the `--bootstrap` path, a root CLI process outside the namespace) is unaffected.

**`xtask/tests/sideload.rs`** holds the script and the unit against each other: the script may not name a path the unit hides, and while it sideloads from a home directory the unit may not set `ProtectHome=`. Neither file can check that on its own — the script does not read the unit, the unit does not know the script. A third test fails if both halves stop applying, so the file cannot go quietly vacuous.

### Notes

- No IPC change and no `API_VERSION` bump: `preflight::Check` is internal, reported through the existing `Error::Preflight`.
- Verified: `cargo test -p updater` (283), `-p xtask`, `-p robotctl` green; `shellcheck scripts/dev-push.sh` clean; the guard test confirmed failing when the default is put back under `/var/tmp`. **Not** verified against the board end to end — it dropped off the network mid-session.
- Stacked on #94, which is where the docs page being edited here comes from.

### DIFF
--- a/docs/design/updater-design.md
+++ b/docs/design/updater-design.md
@@ -996,6 +996,12 @@ Cheap additions that slot into the same IPC / engine and pay off:
   `--version` — an operator naming a directory is not a mirror that has gone
   backwards, and a local build is a prerelease that sorts below whatever the board is
   running, so guarding it would refuse every push.
+  One constraint on the directory, and it is systemd's rather than the engine's:
+  `updaterd.service` sets `PrivateTmp=yes`, which gives the unit its own `/tmp` **and**
+  its own `/var/tmp`, so a release copied to either from a shell is not the one the
+  daemon reads. That is why the sideload directory lives in the board user's home and
+  why `Check::SideloadDir` exists — without it the failure is a missing manifest in a
+  directory whose `ls` lists it, which is unfalsifiable from the outside.
 - **BLE provisioning security.** Adjacent but important: wifi credentials pass
   over BLE during setup. That characteristic must be paired + encrypted, or it's
   a credential leak. Update artifacts are signed so a spoofed *trigger* is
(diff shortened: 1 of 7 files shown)


pierre-rouanet:
Folded into #94 — GitHub marked this merged because its head landed on its base branch, not because it shipped. The fix and the page documenting it are one change about making `dev-push` work, and both commits now sit on #94 against `main`.

PR #116, opened 2026-08-20
robotctl monitor show a cool 3D ascii view of the robot

apirrone:
Capture d’écran du 2026-08-20 09-07-25

Performance wise, the view costs about 2% of one core while you watch a resting robot, bounded at ~6% of one core (≈1.5% of the whole board) when the robot is moving nonstop. 

Hiding the view removes this overhead completely.


pierre-rouanet (approved):
So much love for this!

PR #117, opened 2026-08-20
Skills & behaviors from microduck_runtime (migration step 1)

apirrone:
Migration step 1: the prototype's skills and behaviors, ported to the daemon and shaken down on hardware.

## What this brings (API v9)

- **Skills**: ground pick (A), ball kicks (LB/RB), sit↔stand (DPad-Down), **la roulade** (X — press for one roll, hold to chain), body-pose mode (B), mouth on the triggers, shutdown sit (Select 2 s), seated-boot rise.
- **Roller mode**: `mode = "roller"` in `robotd.toml` — the prototype's rebased roller line (alpha defaults + roller policy, kp 200, scale 0.8, crouch on the ground-pick slot, speeds 0.6/0.5/0.3).
- New calls: `robot.do`, `robot.pose`, `robot.mouth`, `robot.shutdown`, `robot.mode`; `robot.enable` gains `toggle`. `robotctl robot do <skill>` / `robot mode`; monitor shows the loaded skill networks and live skill labels.
- 9 policies vendored (md5-verified against the prototype), shipped in the release; an xtask tripwire keeps all three packaging lists honest.

## Behavior contract

Defaults reproduce the prototype's **current alpha configuration** exactly — same networks, same priority chain (roulade > kick > ground pick > sit/rise > stand > walk), same tuning. Full audit table below; verified end-to-end against the real ONNX on a fake bus, then on the robot.

| value | both |
|---|---|
| kp / ki / kd | 200 / 0 / 0 |
| standing | gain ×0.8 = 160, scale 1.0, threshold 0.05 |
| action scale | walk 0.9, roller 0.8 |
| low-pass head/legs | 0.5 / 0.7 (both modes) |
| ground pick | 4.0 s ×1.0 (roller crouch 3.0 s ×0.8), ends at 70 % |
| kick / rise / roulade | 0.5 s @ standing tuning / 1.0 s / 1.0 s ×1.0 @ kp 200 |
| smoothing / deadzone / speeds | 0.2 / 0.1 / 0.3·0.3·1.5 (roller 0.6·0.5·0.3) |

## Fixed on hardware along the way

- **Servos ran the factory D gain** — the daemon only ever wrote P; the prototype zeroes I/D. Same kP, visibly softer robot. Now written together.
- **`fall → limp` is off by default**, as the prototype is: a fall is a report in the stream, not a gate — no limp, no refused `init`/`enable`/skills. `[safety] fall_limp` (or `fall_recover`, which implies it) arms the old behavior.
- **Start is a robot-side toggle**: the pad kept a local on/off belief that drifted (relax, shutdown, restarts) making Start a dead button every other press. Off returns the robot to its home pose directly; on hands the policy a robot standing at home.
- **The shipped `robotd.toml` pins nothing** — every tuning value is a commented default. A pinned copy from an earlier install froze a board at standing kP 120 while the release said 160.

## Board notes

- Boards installed before this branch should refresh their config once: `sudo rm /etc/robot/robotd.toml && sudo systemctl restart robotd` (defaults are correct; the file is only for overrides).
- API v9: `robotctl` and daemons must come from the same release, which `update apply` guarantees.

Deferred (later steps): autonomous brain, camera/mediad, ToF, audio, IMU-v2 calibration changes.


pierre-rouanet (approved):
Work fine on my robot! Well except is old and not as agile as yours 😂

Just two things:
- Why the whl in the commit? Maybe remove them and force-push to avoid keeping them in history
- And this branch also includes a previous version of PR117. So It conflicts.

And one question:
- When stopping the policy (pressing start on the pad again) the robot directly goes to the init pose. Is that intended?

apirrone:
> Work fine on my robot! Well except is old and not as agile as yours 😂
> 
> Just two things:
> 
> * Why the whl in the commit? Maybe remove them and force-push to avoid keeping them in history
> * And this branch also includes a previous version of PR117. So It conflicts.
> 
> And one question:
> 
> * When stopping the policy (pressing start on the pad again) the robot directly goes to the init pose. Is that intended?

Should be fixed !

PR #125, opened 2026-08-21
Kinematics & odometry: MJCF-driven FK and contact odometry, in the loop

apirrone:
Absorbs the last geometry-owning satellite (`microduck_kinematics_rs`) and ports the runtime's contact odometry, as two library subcrates on one branch. Step 06 of the parity plan — and the blocker for ToF reprojection and gaze IK.

## `kinematics/`

FK parsed from the same `robot_walk.xml` the policies train against: update the MJCF, rerun the fixture generator, and everything downstream still works. No hand-transcribed joint tables anywhere.

Reworked for the hot path rather than transliterated:
- The prototype crate carried joint angles in a `HashMap<String, f64>` and answered one site query by recomputing **every** body into a freshly allocated `Vec`. Here the model is compiled once — names resolved to indices, each site flattened into its own root→site chain — and a query is a zero-allocation fold over ~6 links.
- Measured (release, this laptop, same workload): both feet **882 → 122 ns**, head camera **462 → 47 ns**. A probe stays in `tests/perf_probe.rs` (`--ignored`, prints, asserts nothing).
- The math is a hand-rolled quat/pose pair (~100 lines) instead of nalgebra: the MuJoCo fixture pins it to 1e-6, and what nalgebra would add is compile time, not confidence.

Correctness carried over, not assumed:
- `tests/fk_against_mujoco.rs`: every site on 64 random poses against `mj_kinematics` output (fixture vendored from the satellite repo).
- `head` module: the runtime shim's cv2/sensor convention quats and the alpha sign-pin test (+head_yaw looks left, +head_pitch looks **down**), plus a tripwire that fails if a future MJCF stops mounting the ToF parallel to the camera.
- Alpha only, like the daemon; the parser is generic, so another variant is a dropped-in XML.

## `odometry/`

The Rhoban-derived estimator from the runtime (sole-corner anchoring, IMU re-projection, 2-tick switch confirmation), with two changes:
- The foot chains come from the kinematics model instead of the runtime's hand-copied segment tables — one source of truth for geometry.
- Each foot's chain is evaluated once per tick and its four corners transformed through the result; the runtime re-walked a full leg chain per corner (9 evaluations → 2).

Not a service: a struct `robotd` ticks in its own loop, because its inputs are exactly the sample the loop already read.

## Integration

- `robotd` updates the estimator on fresh samples only (a coasted tick would tell it the robot froze) and only once the SFLP filter is warm (seeding the world origin from a converging quaternion would put it somewhere the robot never was). It runs at the loop's 50 Hz — the runtime's 100 Hz interleave existed to double the bus read rate, which is exactly the pressure #124 taught this loop not to add.
- `robot.state` grows an `odom` field (position + yaw), `#[serde(default)]` like `gravity` — old frames parse, no API bump. `robotctl monitor` shows x/y/yaw on the header's third row.
- No new binaries, no packaging changes.

## Testing

- Workspace tests, clippy, fmt green (44 suites).
- New: MuJoCo fixture parity, planar-arm trig, alpha sign pins, odometry invariants (origin seeding, no standing drift, mouth ignored, anchor switch needs its 2 ticks, continuity under a stance sweep), old-frame compat parse.
- Hardware: install `daemon-dev-kinematics`, drive around, and watch `odom` in the monitor — walking a loop and coming back should read near (0, 0), heading-relative.


apirrone:
Added in cc6cf37: **a top-down map of the odometry track in `robotctl monitor`**, below the 3D robot view. Fixed-size panel; the world zooms out to keep the whole path in frame, with a "N m across" caption. Braille track, `+` at the origin, yellow `●` + heading ray for the robot. Bounded memory (≤2048 points, self-thinning track) and paint-on-redraw only.

```
                 ⢀⡠●⠤⢤⣀
                ⡰⠋ ⢸  ⠈⠳⡀
               ⣸⠁       ⢹⡀
              ⢀⠇         ⢇
              ⢸          ⢸
              ⢸          +
              ⠘⡄         ⡜
               ⢳        ⢠⠃
               ⠈⢣⡀     ⣠⠋
                 ⠙⠢⠤⠤⠤⠚⠁
```

PR #126, opened 2026-08-21
ToF reprojection & gaze IK (robot.look, API v12)

apirrone:
The two consumers #125 unblocked, on one branch as two additions to the `kinematics` crate plus their wiring.

## ToF reprojection — `kinematics::tof::Reprojector`

The prototype's pipeline, rebuilt on the MJCF model:
- The 45°×45° beam table (8×8 zone centres, half-zone inset) rotated through `HeadFk::tof_in_trunk` **per frame**, so a panning head reprojects correctly.
- **Floor filter**: a beam whose slant range × downward component reaches the sensor's height above the floor (×0.85 safety, the prototype's tuned value) is the floor, not an obstacle. The trunk's standing height comes from the MJCF's own `trunk_base` drop height (0.12 m) — the parser now surfaces it (`Model::trunk_height_m`) instead of anyone transcribing a constant.
- **Min-range filter** (0.10 m): under ~10 cm the sensor's cover-glass crosstalk and pile-up produce phantom short returns, so the band is discarded as noise.
- Output per zone: `Empty / TooClose / Floor / Hit { point-in-trunk, horizontal range }` — the classification obstacle avoidance and mapping want.

Placed in `kinematics`, deliberately **not** in the `tof` crate: it's pure geometry, and `tof` carries the vendored ST C driver that clients like `robotctl` must never have to link (the same rule that keeps `robotctl`'s zone parsing independent).

**Visible on hardware**: the monitor's ToF panel runs the reprojection live against the measured head joints and paints floor returns green (legend updated). Pitch the head down at open floor — the grid should go green; put a box in the way — its cells stay on the range ramp.

## Gaze IK — `HeadFk::look_at` and `robot.look`

- `look_at(target_in_trunk, neck_pitch) -> Gaze { joints, clamped }`. The head is a pitch-*then*-yaw gimbal (`head_pitch` is upstream), so the axes are coupled and pitch loses all elevation authority near ±90° yaw — a per-axis solver provably stalls there (tried first, traced, discarded). Solved as damped Gauss-Newton against the real FK with a finite-difference Jacobian: FK is ~50 ns, the whole solve a couple of µs, and the Levenberg damping rides through the singularity.
- The MJCF's joint `range` attributes are now parsed and the solve clamps against them every step; `clamped` reports a target beyond reach (limits or geometry) — the joints are then the closest possible gaze.
- **`robot.look {x, y, z, neck_pitch}`** (API v11 → v12): robotd runs the IK and feeds the existing head-intent path, answering with the exact `HeadParams` it sent — a client holds the gaze by resending them as `robot.head`. This is the gaze form `HeadParams`' doc promised from day one. Routed nowhere over BLE (motor control), refused by updaterd like the rest of `robot.*`.
- **`robotctl robot look 1 0 0`** — X forward, Y left, Z up, metres, trunk frame. `--neck-pitch` sets posture; a clamped answer says "out of reach — this is the closest the head can look".

## Testing

- Workspace green (44 suites), clippy clean.
- IK: FK-verified convergence on targets around the envelope, posture held, behind-the-robot pins at the ±170° yaw limit with `clamped`.
- Reprojection: a metre-out forward return lands a metre ahead; floor at floor distance is `Floor` while the same beam interrupted at 75% is a `Hit`; upward beams are never floor; head yaw steers the points.
- robotd: `robot.look` dispatch test — answer equals exactly what the head intent received, left target → leftward yaw.
- Hardware checklist: `robotctl robot look 1 0 0` then `1 1 0` then `1 0 -0.3` (ahead → left → down); monitor with the head pitched down over open floor (green grid), then with an obstacle (ramp colors). One thing worth confirming on the real sensor: grid left/right orientation — wave a hand on the robot's left and check the floor/hit cells move on the expected side; if mirrored, the beam table's azimuth sign gets one flip.


apirrone:
Hardware shakedown follow-ups (07bf538…d17d89d), all confirmed on the robot:

- **Depth returns in the 3D view** — contact points (yellow = thing, green = floor), not rays; transformed by the visual root's *full* pose (the baked `duck.bin` root keeps its scene rest offset — bare `attitude(gravity)` floats overlays a body-height away). View auto-zooms to frame points within 0.45 m, with hysteresis: out instantly, back in after 5 quiet seconds, 5 cm bins — no breathing at the sensor rate.
- **IMU-aware floor filter** — reprojection takes a `Posture` (projected gravity + odometry's measured trunk height) and computes floor/range verdicts in the gravity-levelled frame. A hand-leaned robot now calls the floor the floor; the prototype's filter had the same blind spot, which likely put gait-correlated phantom obstacles into anything maploc built while walking.
- Sensor pose anchored on the MJCF's own `tof` site (2 cm truer than borrowing the camera's).
- `robotctl robot look` accepts negative coordinates; ← → orbit the 3D view; panel up to 74 columns; `DUCK_MARKERS=demo DUCK_DUMP=… cargo test dump_for_eyeballs` renders the whole marker pipeline for eyeballing without a robot.

PR #131, opened 2026-08-24
media bring-up: GStreamer, hardware H.264, and the camera overlay

pierre-rouanet:
Everything a board needs before `mediad` can exist, proven on a Radxa Zero 3W. It started as an
install script and grew as the hardware answered questions; the findings live in
[`docs/project/media-bringup.md`](docs/project/media-bringup.md).

Nothing here creates `mediad` or touches an existing daemon. It is provisioning, one new script,
and a bug fix in `board-test.sh`.

## What a board gets

**`scripts/setup-gstreamer.sh`** — new. Installs the GStreamer runtime, fetches the prebuilt
plugins at a pinned version, gives the VPU nodes a group, and reports what the hardware can
actually do. Standalone, idempotent, needs no reboot, and persists itself to
`/usr/local/sbin/robot-setup-gstreamer` for re-checking after a kernel change.

**Provisioning installs it by default.** `--no-gstreamer` / `DUCK_GSTREAMER=0` opts out.
Off would have been the safer-looking default and the wrong one: every robot wants this, and a
board provisioned in between would need a bring-up step someone has to remember.

**`configure_camera` in `setup-board.sh`** — enables the head camera's device-tree overlay. The
trick is that Armbian ships it *unprefixed* (`radxa-zero3-rpi-camera-v2.dtbo`) while the board runs
`overlay_prefix=rk3568`, so the loader resolves the word to a file that does not exist, finds
nothing, and boots happily with no camera. The file is mirrored under the prefixed name first.
`DUCK_CAMERA_OVERLAY` selects another module.

**The plugins come from [`pollen-robotics/microduck-gst-plugins`](https://github.com/pollen-robotics/microduck-gst-plugins)**,
a new public repo that builds them natively on an arm64 runner in a `debian:trixie` container and
attaches a tarball to a release. Pinned here at **`v3`**. Public because the download happens during
provisioning and, later, from the updater's `preinstall` hook — which runs with a cleared
environment and no token, exactly like ONNX Runtime already does.

Pinned in `[workspace.metadata.gst-plugins]`, with an `xtask` test asserting the script's literal
agrees — the same arrangement, and the same reason, as `ONNX_VERSION`.

**`v3` carries two patches, both about getting hardware H.264 to a browser**, each documented in
that repo's `patches/README.md` with what it buys and its route upstream:

- `webrtcsink` inserts a software `videoconvert ! videoscale` in front of any encoder it does not
  recognise. `mpph264enc` converts on the SoC's 2D accelerator instead, so that is a full CPU pass
  over every frame to do work the hardware was going to do anyway, on cores `robotd`'s control loop
  shares.
- `mpph264enc`'s src pad template omitted `constrained-baseline` from its profile list — the one
  profile `webrtcsink`'s codec discovery demands. Without it H.264 is silently absent from every
  offer. The bitstream was always correct; only the template denied it, which is measured in the
  findings doc rather than assumed.

## What the hardware settled

| | |
|---|---|
| hardware H.264 | **works** — `mpph264enc`, 1280x720; `h264parse` negotiates `constrained-baseline`, measured, so `profile-level-id=42e01f` is available to WebRTC |
| reached through | Rockchip MPP (`/dev/mpp_service`), **not** V4L2 M2M — the expected shape on a BSP kernel |
| camera | IMX219 probes (`Chip ID 0x0773`), `/dev/video0` up to 3280x2464, frames flowing |
| full chain | sensor → rkisp → mmap capture → NV12 → hardware encode → decodable stream |
| plugin ABI across 1.x | not a risk: a plugin built against GStreamer 1.14 registers cleanly in 1.26.2 |

Four `mpph264enc` properties turn out to be decisions rather than defaults — `profile`,
`header-mode`, `rotation`, `bps` — and two need none: there is no B-frame knob, and the sink pad
takes NV12, which is what rkisp emits.

## Three device nodes need the `video` group, not one

This is the part most worth reading. Each failed differently enough to look like an unrelated bug,
and between them they cost three debugging rounds:

- `/dev/mpp_service` root-only → `mpi_enc_test` writes nothing and **exits 0**, and `mpph264enc`
  is not registered at all. That last one made Radxa's own prebuilt plugin look decode-only when
  it is not: an MPP plugin registers its decoders unconditionally and *probes MPP* before
  registering its encoders.
- `/dev/rga` root-only → the element exists, the pipeline starts, then `rgaCtx=(nil)`.
- `/dev/video0` arrives `root:video` already, so it is the one that does not bite.

`mediad.service` will need `SupplementaryGroups=video` **and**
`Environment=GST_PLUGIN_PATH=/usr/local/lib/gstreamer-1.0` — the plugin directory is deliberately
not the distro's, so `apt` cannot replace it, and it is not searched by default. Two lines standing
between a working pipeline and four distinct confusing failures.

## An unrelated bug this branch exposed

`scripts/board-test.sh` read the workspace version with `grep -m1 '^version' Cargo.toml` — the
first `version =` line anywhere in the file. That worked only because nothing above
`[workspace.package]` happened to carry that key. Adding
`[workspace.metadata.gst-plugins]` made it answer `v1`, and the `board` job failed several minutes
into a release build with `invalid value 'v1' for '--version'`.

Fixed where the query is, with a sed range naming `[workspace.package]` and a refusal if it
resolves to nothing. Reordering my table below `[workspace.package]` would also have made CI pass,
which is exactly why I did not.

## Testing

- **On hardware**, a Radxa Zero 3W: every row of both tables above.
- `setup-gstreamer.sh`'s report branches exercised against fixtures with stubbed `gst-inspect-1.0`
  / `v4l2-ctl` / `stat` — which caught a doubled version string, a GNU-only `\s` in a sed, and
  `--help` failing under `curl | sh` before any of it reached a board.
- `configure_camera` and the udev rule: install, idempotent re-run, direct-apply fallback, missing
  rules directory, and `/dev/rga` absent.
- 38 `xtask` tests pass. shellcheck clean on all four scripts, including the `--shell=sh` and
  `test -x` checks CI runs — `setup-gstreamer.sh` is added to that list and to
  `printed_commands_name_absolute_paths`.

## Known unfinished, and deliberately out of scope

- The encoded bitrate came out ~50× under target on a static green scene. Either CBR collapsing or
  capture below 30 fps; the frame count was never measured. If the latter, the unpinned sensor mode
  is the suspect.
- `rawvideoparse` with a hand-computed `blocksize` is a bring-up shortcut that breaks silently once
  stride padding appears. `mediad` wants its own V4L2 mmap loop into `appsrc`.
- Image quality needs `rkaiq`, which the prototype handles in `radxa_setup/setup_rkaiq.sh`. Until
  then the picture is green and noisy — the ISP on raw defaults, and no reflection on the path.
- The WebRTC design — signalling, token custody, transport authorisation, authority arbitration —
  is the next PR.


pierre-rouanet:
Ran on a Radxa Zero 3W. The default is now **on** (`--no-gstreamer` / `DUCK_GSTREAMER=0` opts out), which was the agreed trigger.

```
GStreamer 1.26.2
webrtcbin present · webrtcsink/webrtcsrc absent
v4l2h264enc absent · x264enc, vp8enc, jpegenc present
kernel 6.1.115-vendor-rk35xx  (vendor BSP)
/dev/mpp_service present
V4L2 devices: none
```

**The encoder question is settled: MPP, not V4L2 M2M.** On a Rockchip BSP kernel the VPU is
exposed through `/dev/mpp_service` rather than as a V4L2 encoder, so the absent `v4l2h264enc` is
the expected shape, not a missing package. The report said that badly and now says it plainly.

Radxa's pool does carry what's needed, by plain `.deb` download — the same route
`microduck_runtime/radxa_setup/setup_rkaiq.sh` already uses on this board:

- `librockchip-mpp1` / `librockchip-mpp-dev` / `rockchip-mpp-demos` at `1.5.0-1_arm64`
- `gstreamer1.0-rockchip1_1.14-4_arm64`

So the MPP branch of the report now names two cheap tests in the right order rather than just
diagnosing: `mpi_enc_test` from the demos package proves the VPU encodes with no GStreamer
involved, and the pool's own plugin is worth a `dpkg -i` before anyone builds one — GStreamer
keeps plugin ABI stable across 1.x, so a bullseye-era build may load into 1.26.2 and make the
source build unnecessary.

Two fixes that came from reading the real output rather than a fixture:

- `V4L2 devices: none` is also exactly what an **unattached camera** looks like — the rkisp
  capture nodes appear only once a sensor is probed. It read as a possible driver fault and said
  nothing either way; it now says so.
- The `v4l2h264enc absent` line could be read as a missing package. The verdict now rules that
  out explicitly.

Still open, and narrower than before: whether we need our own apt repo at all. If the Radxa debs
work as-is, pinning URLs and sha256s the way `setup_rkaiq.sh` does may be enough.

PR #169, opened 2026-08-27
RFC: external per-joint target streaming (robot.setJoints)

cagataycali:
**This is a scoping RFC (design doc only, no implementation yet)** — opening it to get your direction on the shape and the safety design before writing code. The doc lands as `docs/rfc/0001-external-per-joint-write.md`.

## Summary

Add one intent to robotd's `robot.*` socket — **`robot.setJoints`** — that lets an *off-robot* controller stream joint targets at the control rate, plus a matching **`External`** drive mode in which the loop uses those targets in place of the on-device policy. Everything still flows through the existing `duck-control` `Safety` chokepoint; nothing bypasses `Safety::apply`, and `control::Runtime` still holds no IO handle.

## Why

robotd is intent-level by design (`robot.move`/`pose`/`head`/`do` → on-device `Policy` → targets), and that stays the default. But two workflows have no path today:

1. **Run an off-robot policy on the real robot** — a policy too big for the Pi, or one being iterated on a workstation/GPU, can't drive the hardware without re-flashing a `model` component on every change. The teleop / VLA / lerobot / MHS pattern is *controller computes joint targets, streams them at rate*; there's no verb for it.
2. **Bring-up / calibration / trajectory replay** — moving a joint to a commanded angle for a test wants direct targets, not an intent the policy reinterprets.

This is **distinct from deploying a finished on-device policy**, which the updater's `model` component already serves (`update.apply`, gated by `robot.modelApi`). The RFC does not touch that path.

## The safety thesis

`robot.setJoints` must be *no less safe* than the policy path, and one gap has to close. `duck-control/src/safety.rs` (lines 42–46) says the range clamp is the *actuator's* travel, **not** a per-joint anatomical limit — "the real joint limits live in the MJCF, which is not vendored here." A policy trained in that MJCF stays inside those limits implicitly; **arbitrary external targets do not.** So, all inside the existing `Safety` chokepoint:

- **mode + torque gate** — `setJoints` honoured only in `External` mode with torque on; refused otherwise (named error), like a refused `run_policy`.
- **external deadman** — reuse the deadman mechanism on a dedicated external-target clock; a stalled off-robot controller must not leave a live command.
- **per-tick step clamp (new)** — bound `|target[j] − previous[j]|` to a max joint velocity so one bad frame can't snap a joint.
- **per-joint anatomical limits (new)** — vendor per-joint `[min,max]` from the MJCF and clamp to them, closing the gap above.
- **NaN refusal** — already present.

## Shape (details in the doc)

- `duck-ipc-proto`: `method::ROBOT_SET_JOINTS = "robot.setJoints"` + `Call::RobotSetJoints(SetJointsParams { targets: Vec<f64> /* len NUM_JOINTS */, gain: Option<u16> })`, a notification like `robot.move`, with a serde round-trip + exact-wire test.
- robotd: `External` drive mode via `robot.setMode`; external-target slot + own age clock in `intents.rs`; loop becomes `read → observe → gate → {External ? external_targets : policy} → safety.apply`.
- Downstream (separate PR): upgrades the `strands-robots` native driver from delegate-only to a real passthrough mode.

## Open questions for you

1. Absolute radians (RFC's proposal — easiest to clamp) vs. offsets from `DEFAULT_POSITION`?
2. A new `External` mode, or gate `setJoints` on an existing mode?
3. Per-joint limits: vendor from the MJCF into `duck-control`, or a config file robotd loads?
4. External deadman: reuse 500 ms, or a tighter dedicated value for rate-streamed control?
5. Is the mouth (slot 9) writable via `setJoints`, or masked like the policy path masks it?

Happy to implement along whichever lines you prefer.

### DIFF
--- a/btd/src/route.rs
+++ b/btd/src/route.rs
@@ -205,6 +205,13 @@ fn permits(call: &proto::Call) -> bool {
         RobotMove(_) | RobotHead(_) | RobotLook(_) | RobotEnable(_) | RobotDo(_) | RobotPose(_)
         | RobotMouth(_) => false,
 
+        // Raw per-joint targets streamed from an off-robot controller (the External drive path).
+        // This is the most direct motor control there is — it bypasses the policy and commands the
+        // servos straight — so it is refused over BLE for every reason `robot.move` is, and then
+        // some: a 20-byte, ~73-s-late notification link is exactly the wrong transport for a joint
+        // stream. External drive belongs on the local socket / a trusted controller over WebRTC.
+        RobotSetJoints(_) => false,
+
         // Harmless and rather charming from a phone — but it rides the same refusal as the
         // rest of robot.* until the app path exists to want it: opening one call to the
         // radio ahead of a client that can use it buys nothing and widens the surface.
--- a/docs/design/robotd-design.md
+++ b/docs/design/robotd-design.md
@@ -552,6 +552,43 @@ that is what the deadman reads.
 `look` (gaze direction) is deferred; both gaze forms will be exposed, and arbitration between
 them is last-writer-wins with no blending.
 
+#### External drive — `robot.setMode "external"` + `robot.setJoints`
+
+The intents above ask the *policy* to walk, look and pose; **External drive** takes the policy
+out of the loop entirely and lets an off-robot controller command the servos directly. It is the
+path a mid-training RL policy or a teleoperation rig takes to drive the physical robot without
+running on it.
+
+```jsonc
+{"jsonrpc":"2.0","id":9,"method":"robot.setMode","params":{"mode":"external"}}
+// absolute joint angles, radians, in JOINT_NAMES order; length must equal NUM_JOINTS
+{"jsonrpc":"2.0","id":10,"method":"robot.setJoints","params":{"targets":[/*15*/],"gain":180}}
+```
+
+Three properties make this safe to expose:
+
+- **It is a drive *source*, not a policy mode.** `external` sits alongside `walk`/`roller` in
+  `robot.setMode`, but it loads no policy and does no homing — the tick *is* the supplied targets.
+  `robot.setJoints` enters it implicitly (a single frame both supplies targets and takes the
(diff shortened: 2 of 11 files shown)


pierre-rouanet:
Hi @cagataycali!

I like your suggestion, Yet from our internal testing, the timing for motor control should be really accurate. Otherwise, behavior just won't work.
So, except if we manage to train policy more robust to this kind of changes, I'm afraid this will not be usable in real scenario.

pierre-rouanet:
Closing this on the answer already given above, and putting the reason somewhere it will
outlive the thread.

`robot.setJoints` is not a shape this robot can serve today. The timing for motor control
has to be accurate enough that the policy's own assumptions hold, and a controller
streaming per-joint targets from off the robot cannot meet that over a socket — the jitter
alone breaks the behaviour, before any question of bandwidth or safety. That is a
property of the policies we have rather than of the API you designed, so this is a "not
now" and not a "never": if we train a policy robust to it, the RFC is the right starting
point and can be reopened as it stands.

Worth saying what the RFC got right, because it is not obvious from a close. Routing
everything through `Safety::apply` rather than beside it is the correct thesis, and
catching that `duck-control/src/safety.rs` clamps to the *actuator's* travel rather than a
per-joint anatomical limit — and that an external writer is exactly the caller who would
find that gap — is a real observation about the current code, independent of whether this
verb ever ships.

On the two workflows that motivated it, to be straight rather than reassuring: neither has
a path today. `robot.pose` is body height, roll and pitch — not a per-joint target — so
moving one joint to a commanded angle for a bench test genuinely has no verb either. That
one is a much smaller ask than streaming at the control rate, and it is not subject to the
timing objection, so it is worth its own issue if you want it. The off-robot policy case is
the one to describe as a workflow rather than as a verb, since the answer there may well
not be a per-joint write at the loop rate at all.

Thanks for opening it as an RFC first rather than as code — that was the right call and it
made this a cheap conversation.

Issue #153, opened 2026-08-26
`robotctl quack` should open the mouth too


apirrone:
Easier to know which robot quacks when multiple are close

PR #170, opened 2026-08-28
Shutdown must not stand the robot back up, and quack opens the beak

### LINKED ISSUE #159: [bug] when shutting down, it sits down (normal) but then goes to init (the bug)

sankalpsthakur:
Fixes #159 and #153.

Prevent a completed shutdown from re-enabling torque through the same tick's stale enable snapshot. Ignore `robot.init` while shutdown is in progress or power is off, skip enable-driven bring-up during that sequence and set `Bringup::Limp` on the no-sit power-off path.

`robotctl quack` now holds the beak open for 450 ms, including when audio playback fails. It preserves the blocking goodbye behavior and avoids taking over while ride, theremin or chorale owns the PCM. The mouth can move while sitting but torqued.

Motor-gain tuning is unchanged: policy.gain remains 200 with I/D at zero.

Also remove a duplicate `#[cfg(test)]` attribute that failed strict Clippy.

## Test plan

- [x] `cargo test -p robotd`: 99 unit tests and 7 updater_gate tests passed, including shutdown/init and one-shot beak regressions.
- [x] `RUSTFLAGS="-D warnings" cargo clippy -p robotd --all-targets`
- [ ] On a real duck: shutdown sits and stays down; init cannot stand it up before a new power-on.
- [ ] On a torqued robot: `robotctl quack` opens the beak and plays the call.

The tested shutdown path has no policy and powers off immediately. The sit-with-policy path requires ONNX and remains unvalidated.

Codex assisted with this description.

### DIFF
--- a/docs/robot/cheatsheet.md
+++ b/docs/robot/cheatsheet.md
@@ -558,7 +558,7 @@ robotctl quack
 
 The loudest way to tell ducks apart: every robot's voice bank is generated from its SoC
 serial (`sounds ensure-bank`, run by every release install), so the robot that answers — in
-a voice that is only its own — is the one you're SSH'd into. A robot with no voice — audio
+a voice that is only its own, beak open — is the one you're SSH'd into. A robot with no voice — audio
 off, or no bank — says so instead of printing 🦆, so silence always means the wrong duck.
 The robot also greets when `robotd` comes up, pecks goodbye before powering off, and — if you
 ask it to — coos when the mic hears its head being scratched. That one is off by default in
--- a/robotd/src/chorale.rs
+++ b/robotd/src/chorale.rs
@@ -542,7 +542,7 @@ impl Chorale {
         // holding the beat, and two beacons carrying a piece would be two conductors.
         let idle = self.beacon(proto::ChoraleBeacon::IDLE, self.heartbeat(now), Vec::new());
         let advertise = self.publish(Some(idle), true);
-        let singing = position.and_then(|at| self.my_part(&roster).map(|part| (part, at)));
+        let singing = self.my_part(&roster).zip(position);
         Tick {
             advertise,
             joining: singing.is_none(),
@@ -630,7 +630,6 @@ pub fn head_expression(beats: f64, reach: f64) -> [f64; 4] {
     ]
 }
 
-#[cfg(test)]
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/robotd/src/main.rs
+++ b/robotd/src/main.rs
@@ -1832,11 +1832,13 @@ async fn control_loop<T: RobotIo>(
         // still-set `enabled` flag wins — `request_relax` clears that flag, and reading the request
         // first means the order cannot invert.
         match intents.take_power_request() {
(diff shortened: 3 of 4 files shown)


pierre-rouanet:
Closing this one because it has been overtaken on one half and outgrown on the other —
not because the diagnosis was wrong. It was right on both counts.

**The shutdown half has landed independently.** `7563b4c` ("A reset stood the robot up, and
a shutdown stood it up again") fixed the stale-enable-snapshot path on 2026-09-02, five
days after you opened this, and #319 today closed the last related door — a mode switch
that was waiting when the robot goes down is now dropped rather than standing it back up
out of the sit. `robot.init` is ignored while `powered_off`, with a test. So there is
nothing left to take there.

**The beak half is still wanted, and #153 is still open.** `robotctl quack` sends
`robot.sound` with a `Chirp` and nothing moves the mouth. Your instinct to put it in
`robotd/src/sound.rs` rather than in `robotctl` is the right one and I want that on the
record: holding the beak for the duration of the call means every caller of `robot.sound`
gets it — the phone, a pad binding, a script — instead of only the one CLI subcommand. The
care about not taking the PCM while ride, theremin or chorale owns it is the right worry
too.

What I am not asking for is a rebase of *this* branch. `robotd/src/main.rs` has been
substantially rewritten underneath it since August — the bring-up state machine, the
shutdown sit and the chorale have all moved — and untangling the beak change out of a
four-file diff whose other half is already fixed is more work than writing it fresh against
current `main`.

So: a new PR, `sound.rs` only, beak open for the length of the call, closing #153, would
be very welcome and I will review it quickly. Sorry this sat as long as it did.

sankalpsthakur:
> Closing this one because it has been overtaken on one half and outgrown on the other —
> not because the diagnosis was wrong. It was right on both counts.
> 
> **The shutdown half has landed independently.** `7563b4c` ("A reset stood the robot up, and
> a shutdown stood it up again") fixed the stale-enable-snapshot path on 2026-09-02, five
> days after you opened this, and #319 today closed the last related door — a mode switch
> that was waiting when the robot goes down is now dropped rather than standing it back up
> out of the sit. `robot.init` is ignored while `powered_off`, with a test. So there is
> nothing left to take there.
> 
> **The beak half is still wanted, and #153 is still open.** `robotctl quack` sends
> `robot.sound` with a `Chirp` and nothing moves the mouth. Your instinct to put it in
> `robotd/src/sound.rs` rather than in `robotctl` is the right one and I want that on the
> record: holding the beak for the duration of the call means every caller of `robot.sound`
> gets it — the phone, a pad binding, a script — instead of only the one CLI subcommand. The
> care about not taking the PCM while ride, theremin or chorale owns it is the right worry
> too.
> 
> What I am not asking for is a rebase of *this* branch. `robotd/src/main.rs` has been
> substantially rewritten underneath it since August — the bring-up state machine, the
> shutdown sit and the chorale have all moved — and untangling the beak change out of a
> four-file diff whose other half is already fixed is more work than writing it fresh against
> current `main`.
> 
> So: a new PR, `sound.rs` only, beak open for the length of the call, closing #153, would
> be very welcome and I will review it quickly. Sorry this sat as long as it did.

np i understand

PR #179, opened 2026-08-31
auto-exposure: write the digital gain back down to 1x

### LINKED ISSUE #180: auto-exposure never writes the digital gain back down to 1x
in `mediad/src/exposure.rs`, `write()` only sends the `gain` control when it is above 256, and 256 is 1x. so the one step that takes the isp back down to 1x gets skipped along with the no-ops.

a dark room pushes the digital gain up. when the light comes back the loop walks it down and eventually asks for 1x, and that write never goes out, so the isp keeps the last value above 1x it was given. after that the loop is metering a picture that is still amplified, reads it as too bright, and shortens the shutter to make up the difference, which is the control the module tries to spend first.

it survives a restart too. `pipeline.rs` pins `exposure` and `analogue_gain` through extra-controls but nothing pins `gain`, and `Ae::starting_at` always starts at 1.0 digital, so a `mediad` that comes back after an update has no idea what the isp is actually holding.

i stepped `Ae` from luma 1 up to luma 250 to check. it stops telling the isp anything once the gain gets back to 256, and ends up with the shutter at 4 lines while the isp is still at 1.375x.

seen on 0.10.0, main at 590b986.

Nixxx19:
fixes #180

write() only sends the gain control when it's above 1x, and 256 is 1x, so the one step that takes the isp back down gets dropped along with the no-ops. dark room puts the gain up, light comes back, the loop walks it down to 1x and that write never goes out.

so the picture stays amplified while the loop thinks it isn't, and it pays for the difference out of the shutter, which is the control it tries to spend first.

now it tracks whether the last write left the gain up and sends it if so. a node with no digital gain still never gets a gain write, which is what the original skip was there for. two tests, both fail on main.

### DIFF
--- a/mediad/src/exposure.rs
+++ b/mediad/src/exposure.rs
@@ -229,8 +229,8 @@ pub fn spawn(device: String, frames: Frames, exposure_lines: u32, analogue_gain_
         .spawn(move || {
             let mut ae = Ae::starting_at(exposure_lines, analogue_gain_reg);
             let mut proven = false;
-            // whether the isp is holding a digital gain above 1x. see `worth_writing_gain`.
-            let mut raised = false;
+            // What this run knows about the ISP's digital gain. See `Digital`.
+            let mut digital = Digital::Unknown;
             let mut waiting_logged = false;
             // Ticks since the last write, for the re-assert heartbeat.
             let mut quiet = 0u32;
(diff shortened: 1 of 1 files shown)


Nixxx19:
hi @pierre-rouanet, could you take a look at this one when you have a minute? it stops the camera getting stuck brighter than the loop thinks it is.

PR #181, opened 2026-09-01
apply clippy float optimization recommendations

Lucas-BRT:
This PR applyes the following clippy nursery recomendations:

https://rust-lang.github.io/rust-clippy/rust-1.96.0/index.html#imprecise_flops
https://rust-lang.github.io/rust-clippy/rust-1.96.0/index.html#suboptimal_flops

Some of then can affect the readability (specially for the mathematical point of view), but the speed gain can have some benefit in the final product. i don't have the hardware to test, so i can't measure the exact speed gains in a specific processor in normal conditions, i will be grateful if someone can test if this have some real gains or we should only add a #[allow] for the linter.

Also, not related to this PR in specific, but in some places the codebase triggers other Linter warnings, like using the Struct name instead of Self, is some specific reason for this or i can open a PR to fix?

### DIFF
--- a/kinematics/src/head.rs
+++ b/kinematics/src/head.rs
@@ -141,7 +141,7 @@ impl HeadFk {
             ];
             // cv2 camera axes: +x right, +y down, +z forward.
             let v = cam.quat.conjugate().rotate(v);
-            let flat = (v[0] * v[0] + v[2] * v[2]).sqrt();
+            let flat = v[0].hypot(v[2]);
             [v[0].atan2(v[2]), v[1].atan2(flat)]
         };
 
--- a/kinematics/src/head.rs
+++ b/kinematics/src/head.rs
@@ -166,19 +166,22 @@ impl HeadFk {
 
             // Solve (JᵀJ + λI) Δ = -Jᵀe, the 2×2 case written out.
             let (a, b) = (
-                j[0][0] * j[0][0] + j[1][0] * j[1][0] + LAMBDA,
-                j[0][0] * j[0][1] + j[1][0] * j[1][1],
+                j[1][0].mul_add(j[1][0], j[0][0] * j[0][0]) + LAMBDA,
+                j[1][0].mul_add(j[1][1], j[0][0] * j[0][1]),
             );
-            let d = j[0][1] * j[0][1] + j[1][1] * j[1][1] + LAMBDA;
+            let d = j[1][1].mul_add(j[1][1], j[0][1] * j[0][1]) + LAMBDA;
             let g = [
-                j[0][0] * e[0] + j[1][0] * e[1],
-                j[0][1] * e[0] + j[1][1] * e[1],
+                j[1][0].mul_add(e[1], j[0][0] * e[0]),
+                j[1][1].mul_add(e[1], j[0][1] * e[0]),
             ];
-            let det = a * d - b * b;
+            let det = b.mul_add(-b, a * d);
             if det.abs() < 1e-12 {
                 break; // fully singular and damped-out: nothing left to gain
             }
-            let step = [(-d * g[0] + b * g[1]) / det, (b * g[0] - a * g[1]) / det];
+            let step = [
+                b.mul_add(g[1], -d * g[0]) / det,
+                a.mul_add(-g[1], b * g[0]) / det,
+            ];
             let scale = (MAX_STEP / step[0].hypot(step[1])).min(1.0);
             joints[1] = clamp(joints[1] + scale * step[0], range(1));
             joints[2] = clamp(joints[2] + scale * step[1], range(2));
--- a/kinematics/src/math.rs
+++ b/kinematics/src/math.rs
@@ -40,7 +40,14 @@ impl Quat {
     /// attribute, and a model that ignores a broken quat is diagnosable where a
     /// model full of NaN is not.
     pub fn normalized(self) -> Self {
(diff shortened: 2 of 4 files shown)


pierre-rouanet:
We ran this on the real board, so here are the numbers you asked for.

**Setup**: Radxa Zero 3 (RK3566, 4× Cortex-A55 @ 1.8 GHz), Armbian 26.8.1 trixie, core pinned with `taskset`, 1.8 GHz and 68 °C held for the whole run — no thermal throttling. Built with the repo's own `cargo zigbuild --release --target aarch64-unknown-linux-gnu.2.31`, rustc 1.97.1, stock release profile (no LTO, no `target-cpu`), i.e. the shape a release actually ships.

Three builds, compiled identically against one bench binary driving the public API with deterministic inputs: the merge base (`9f7eaad`), this PR's head (`74e265b`), and a third that is **this PR with only its two new `hypot` calls reverted, every `mul_add` kept**. The three run alternately, 5 passes each, 15 timed rounds per case, fastest reported — the estimator least polluted by the `robotd`/`mediad` load that is on the board anyway.

ns per call, lower is better:

| case | base | this PR | vs base | no-`hypot` | vs base |
|---|---|---|---|---|---|
| `quat_mul` | 34.57 | 33.46 | −3.2% | 32.92 | −4.8% |
| `quat_rotate` | 42.09 | 32.59 | **−22.6%** | 32.40 | −23.0% |
| `quat_normalized` | 64.41 | 63.91 | −0.8% | 63.90 | −0.8% |
| `quat_yaw` | 128.66 | 128.15 | −0.4% | 128.13 | −0.4% |
| `fk_site_pose` | 912.05 | 797.11 | **−12.6%** | 795.97 | −12.7% |
| `head_camera_fk` | 824.21 | 710.33 | **−13.8%** | 710.18 | −13.8% |
| `head_look_at` | 9151.49 | 8466.17 | −7.5% | 8017.18 | **−12.4%** |
| `tof_project` | 6573.86 | 7786.31 | **+18.4%** | 5728.40 | **−12.9%** |

## The two rewrites pull in opposite directions

**`mul_add` pays.** It lowers to a single `fmadd` on aarch64 — the baseline binary contains zero FMA instructions, this PR's contains 181. The compiler will not contract `a * b + c` on its own, because that changes the rounding, so this is a change only a human or a lint can make. It buys 12–14% off forward kinematics and 23% off `Quat::rotate`, which is more than we expected on an in-order A55.

**`hypot` costs.** It does *not* lower to `fsqrt`. It stays an out-of-line call into libm (`bl <hypot>` in the disassembly), doing exponent-scaling work to avoid intermediate overflow that this code does not need. This PR added two:

- `head.rs:144` — `v[0].hypot(v[2])`, inside `look_at`'s error closure, so three calls per Gauss-Newton iteration via the finite differences.
- `tof.rs:186` — `dir_level[0].hypot(dir_level[1])`, once per zone, so 64 per ToF frame.

The ToF one alone is what turns a 12.9% win into an 18.4% regression on `tof_project`. Reverting just those two lines while keeping every `mul_add` is the last column, and it is the best build in every single case.

## Accuracy

Result checksums are bit-identical between base and this PR everywhere except `head_look_at`, which differs in the 12th significant digit (`8.245243734147e4` → `8.245243734136e4`) accumulated over 20k solves. That is FMA's single rounding, so it is *more* accurate, not less. All 28 `kinematics` tests pass on this PR's head, MuJoCo fixture comparison at 1e-6 included.

## What we would suggest

To be straight about magnitude first: these are microseconds inside a 20 ms control tick. Saving 115 ns on an FK query does not change whether the robot walks. So the gain is real and measurable — that answers your question — but it leaves the readability trade-off where it already was, a judgement call rather than a performance argument that settles itself.

1. **Revert the two new `hypot` calls** (`head.rs:144`, `tof.rs:186`) to their `(a * a + b * b).sqrt()` form. This is the one part that is not a judgement call — it is a straight regression on the hottest path in the diff. Note `head.rs:185` already used `hypot` before this PR; it is on a cold path so we left it alone in the measurement, but it is the same trap.
2. **Keep the `mul_add` rewrites** where they still read acceptably.
3. Where the rewrite mangles a recognisable formula, a module-level `#[allow]` is the better trade. `Quat::mul` is the clearest case: it reads considerably worse and gained 3%. `Quat::rotate` gained 23% and is worth it.

On your second question — `Struct` instead of `Self` — there is no deliberate reason, so a PR for that is welcome.

pierre-rouanet (changes requested):
See comments above.

pierre-rouanet:
Closing this as stale rather than as wrong — and I want to be clear about which, because
the `mul_add` half of it is a real win that we measured and would still take.

Where it stands: the review above named one thing that is not a judgement call — the two
new `hypot` calls at `head.rs:144` and `tof.rs:186` are a straight regression on the
hottest path in the diff, and reverting just those two lines while keeping every
`mul_add` was the best build in every case we timed. There has been no push in three
weeks, and `main` has moved underneath the branch since.

A fresh PR is welcome and would go in: the `mul_add` rewrites, without the two `hypot`
calls, and with a module-level `#[allow]` where the rewrite mangles a recognisable
formula — `Quat::mul` being the clearest case, since it reads considerably worse for 3%.
`Quat::rotate` at 23% is worth the readability cost.

And the answer to your second question stands: there is no deliberate reason for
`Struct` over `Self`, so a PR for that is welcome whenever you feel like it.

Thanks for pushing on this — the question you asked is what got the board benchmark
written, and the FMA finding is one we would not otherwise have had.

PR #187, opened 2026-09-01
padd: refuse --hz 0 instead of dividing by it

### LINKED ISSUE #184: padd --hz 0 panics at startup instead of refusing the flag
`padd --hz 0` panics on startup instead of saying the flag is wrong.

`--hz` is a plain `u32` with no lower bound, and `padd/src/main.rs` builds the tick period as `1.0 / args.hz as f64`. at zero that is infinity, and `Duration::from_secs_f64` panics on infinity rather than giving back a very long period:

thread 'main' panicked at library/core/src/time.rs: cannot convert float seconds to Duration: value is either too big or NaN

it lands late, after the gamepad subsystem is up and after the socket to robotd is connected, so what an operator sees is a daemon that looked like it was starting and then died with a backtrace. systemd restarts it and it dies the same way again.

every other rate in the repo already guards this. `tof` uses `hz.max(1)`, `mediad`'s detector uses `hz.max(0.1)`, `robotctl monitor` clamps to 1..200, and `robot.subscribe` filters zero out before dividing. padd is the only one that does not.

on 0.10.0, main at 590b986.

Nixxx19:
fixes #184

`--hz` was a plain `u32`, so `--hz 0` reached `Duration::from_secs_f64(1.0 / 0.0)` and panicked once the daemon had already brought up the pad and connected to robotd.

this bounds the flag at 1 in clap, so it is refused at parse time with a message naming the flag rather than a backtrace. refused rather than clamped, because a rate quietly changed to something else is one nobody can explain later.

the test parses the flag both ways and fails on the old attribute.

`cargo test -p padd` and `cargo test --workspace` are green.

### DIFF
--- a/padd/src/main.rs
+++ b/padd/src/main.rs
@@ -111,10 +111,13 @@ struct Args {
     /// How often to send intents. Matching the control rate exactly buys nothing — the loop
     /// reads the latest value once per tick — but staying at or above it keeps the added
     /// latency under one tick.
-    ///
-    /// refused below 1 rather than clamped: zero reaches `1.0 / 0.0` below, and
-    /// `Duration::from_secs_f64` panics on infinity instead of producing a very long period.
-    #[arg(long, default_value_t = 50, value_parser = clap::value_parser!(u32).range(1..))]
+    // Bounded both ways, and refused rather than clamped so the flag says what it did.
+    // Zero reaches `1.0 / 0.0` and `Duration::from_secs_f64` panics on infinity. The ceiling
+    // is the other half of the same line: a rate this loop cannot keep gives a period of 0 ns,
+    // `checked_sub` never has anything left to sleep on, and the pad spins on robotd's socket —
+    // the same busy loop `robotctl monitor` clamps for, and the range `control.hz` already
+    // rejects outside.
+    #[arg(long, default_value_t = 50, value_parser = clap::value_parser!(u32).range(1..=1000))]
     hz: u32,
 
     /// Deflection below this counts as centre. Analogue sticks rarely rest at exactly zero,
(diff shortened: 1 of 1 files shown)


pierre-rouanet (commented):
Built the branch: `cargo test -p padd` passes, fmt and clippy clean, and `--hz 0` now fails at parse time naming the flag. Right mechanism — refusing beats clamping here. Two things before merge.

**The bound is one-sided, and the other side is the same line.** `--hz 4294967295` gives `from_secs_f64(1.0 / 4294967295.0)` = 0ns, so `period.checked_sub(tick.elapsed())` at `padd/src/main.rs:611` is always `None`, nothing sleeps, and the loop hammers robotd's socket flat out. That's a worse failure than the panic it replaces: the panic is loud in the journal, a spin loop is a quietly pegged core. The repo already decided this two-sided twice — `robotd-params/src/lib.rs:974` rejects `control.hz` outside `1..=1000`, and `robotctl/src/monitor.rs:814` clamps `1..200` with the comment "`--hz 1000` must not turn this into a busy loop." `tof` only escapes because its `hz` is `u8`. `range(1..=1000)` costs nothing extra, and it also fixes the message, which currently reads `0 is not in 1..=4294967295` — advertising 4294967295 as a valid rate.

**The rationale is now in operator-facing `--help`.** The blank `///` line splits short from long help, so `padd --help` prints "`Duration::from_secs_f64` panics on infinity" to whoever is driving the robot — nothing they can act on, and clap already states the real constraint on error. That note is for whoever next edits the line, so make it a plain `//` above the attribute. Same spot: every other arg doc in the struct starts capitalised.

The test is at the right seam (the period is computed inline in `main`, so parse-time is the only one) and the three assertions are the correct three; the only gap is a case at the top of the range, which follows from the first point.

pierre-rouanet (changes requested):
Marking this as changes-requested for the two points in my review above — the ceiling on `--hz` and the rationale in `--help`. The diagnosis and the mechanism are right; happy to approve once those land.

Nixxx19:
both done, thanks.

the ceiling was the half i missed, and the reason you give is the one that decides it: a panic is loud and a spin loop is not. it is now `range(1..=1000)`, matching what `control.hz` already rejects outside, and the error reads `0 is not in 1..=1000` rather than advertising the top of a u32 as a rate.

the rationale moved to a plain `//` above the attribute, so `--help` is back to the sentence an operator can act on, and clap states the constraint on error. capitalised to match the rest of the struct.

added the two cases at the top of the range, and one at 4294967295 with a note saying why that one is a spin loop rather than a slow tick.

fmt, clippy and `cargo test -p padd` are green.

pierre-rouanet (approved):
The fix is right, and the reasoning in the comments holds up against the code.

`--hz 0` reached the divide at `padd/src/main.rs:254`, which is *after* the pad is up and the
robotd connection is made — so it was a backtrace out of a running daemon, not a parse error.
Moving it into clap is the right place for it.

The ceiling earns its place too: at the top of a `u32` the period rounds to 0 ns, `checked_sub`
at line 611 never has anything left, and the loop spins on robotd's socket unpaced. That is the
quieter of the two failures, so bounding both ends is the better fix.

`1..=1000` is not an arbitrary pick — it is the range `robotd-params/src/lib.rs:974` already
rejects `control.hz` outside, and `robotctl monitor` clamps for the same busy loop
(`robotctl/src/monitor.rs:815`). Refusing rather than clamping is the same choice `control.hz`
made; `monitor` can clamp because its rate is a display rate nobody meters.

I checked the other rate divides in the tree and `padd` was the only unguarded one:
`tof/src/main.rs:273` has `hz.max(1)`, `mediad/src/detect.rs:170` has `hz.max(0.1)`,
`robotd/src/main.rs:2607` filters `> 0`, and `duck-bench` treats `--hz 0` as deliberate unpacing.
So this closes the class, not just the one report.

One thing to add: clap does not render a `value_parser` range in `--help`, only in the error you
get when you trip it, and the doc comment above `hz` still reads as though any value goes. Worth
saying the range where `padd --help` will show it:

```rust
/// How often to send intents, 1–1000 Hz. Matching the control rate exactly buys nothing —
```

Nixxx19:
added, your line verbatim. `padd --help` now opens with "How often to send intents, 1–1000 Hz", so the bound is visible without having to trip it first.

good point that the value_parser range only ever surfaces in the error. i had assumed clap put it in the help text and never checked, which is the same mistake as leaving the rationale in there.

checked it renders: `padd --help` shows the range, and `--hz 1001` gives `1001 is not in 1..=1000`.

fmt, clippy and `cargo test -p padd` still green.

pierre-rouanet:
thanks @Nixxx19 — main moved under this one: #204 rewrote the `--hz` doc comment, since the flag now paces the pad read and intents are held back from there. i resolved that against your branch, but couldn't push to the fork from here, so your three commits are in #208 unchanged, with the merge on top. closing this in favour of that one — the change and the authorship are yours.

PR #188, opened 2026-09-01
duck-detect: keep the c_char cast, which clippy only sees half of

### LINKED ISSUE #185: clippy for the board's target fails on duck-detect, and its suggestion breaks the host build
linting the workspace for the board's own target fails on `duck-detect`, and taking clippy's advice breaks the host build.

`c_str` in `duck-detect/src/rknn.rs` maps a `&[c_char]` with `*c as u8`. `c_char` is `i8` on x86_64 and `u8` on aarch64, so on the board's target that reads as `u8 as u8` and `unnecessary_cast` fires. with `RUSTFLAGS=-D warnings`, which is what CI sets and what CONTRIBUTING's own lint line implies, it is an error rather than a warning:

error: casting to the same type is unnecessary (`u8` -> `u8`), duck-detect/src/rknn.rs:401

and applying the suggestion is worse than leaving it. drop the cast and `cargo check --target x86_64-unknown-linux-gnu` stops compiling with "a value of type `Vec<u8>` cannot be built from an iterator over elements of type `i8`". so the lint asks for a change that breaks the target CI actually builds.

CI does not see it today because the `check` job runs on x86_64. it bites whoever follows the CONTRIBUTING line about linting against the board's target, which is the line that exists because "lint against the board's target or the breakage ships".

on 0.10.0, main at 590b986.

Nixxx19:
fixes #185

`c_char` is `i8` on x86_64 and `u8` on aarch64, so the cast in `c_str` is load bearing on the host and dead on the board, and clippy only ever sees one of the two. linting for the board's target turned it into an error under `-D warnings`, and removing the cast stops the host compiling.

so the cast stays and the lint is allowed on that one function, with the reason written down next to it. nothing else was touched.

checked both ways: `cargo clippy --workspace --all-targets` on aarch64 is clean, and `cargo check -p duck-detect --target x86_64-unknown-linux-gnu` still builds.

### DIFF
--- a/duck-detect/src/rknn.rs
+++ b/duck-detect/src/rknn.rs
@@ -394,6 +394,12 @@ impl Drop for Model {
 }
 
 /// A NUL-terminated vendor string, without the noise.
+///
+/// the cast is needed and clippy cannot see it. `c_char` is `i8` on the x86_64 host and `u8` on
+/// the board, so on the board this is `u8 as u8` and `unnecessary_cast` fires, while taking the
+/// cast out stops the host compiling at all. lint the workspace for the board's own target, the
+/// way CONTRIBUTING asks, and `-D warnings` turns that into an error.
+#[allow(clippy::unnecessary_cast)]
 fn c_str(raw: &[c_char]) -> String {
     let bytes: Vec<u8> = raw
         .iter()


pierre-rouanet (approved):

pierre-rouanet:
Thanks @Nixxx19

PR #189, opened 2026-09-01
ci: lint the scripts that run on a board, not just some of them

### LINKED ISSUE #186: ci does not shellcheck the hooks or the scripts that run on every boot
the shellcheck step in `ci.yml` skips several scripts that run on a board, including the two the workflow's own comment argues hardest for.

the list has 12 names on it. these are not among them:

`hooks/postinstall`, which CONTRIBUTING calls the only thing that runs on every board on every update. `hooks/preinstall.in`, which is rendered into the hook that runs before that. `scripts/robot-boot-check`, which runs on every boot. `scripts/setup-npu.sh` and `scripts/setup-rkaiq.sh`, both fetched standalone with curl and both run by the update hook, exactly like `setup-gstreamer.sh`, which is on the list.

the reason the workflow gives for linting `robot-rescue` is that it is "the script that runs when a board cannot start its update daemon, the worst possible moment to find a typo, and it ships as text like the rest". that argument covers every file above word for word.

all of them are clean right now, which is what makes this cheap to close. the gap is that nothing stops the next edit to any of them shipping a typo to every board.

on 0.10.0, main at 590b986.

Nixxx19:
fixes #186

adds the five that were missing to the shellcheck loop: `setup-npu.sh`, `setup-rkaiq.sh`, `robot-boot-check`, `hooks/postinstall`, and `hooks/preinstall.in`.

the template gets its own step because it is the one file here that is not executable. `xtask package` renders it into the hook, so the file in the tree is not the file that runs and `test -x` would fail on it. its placeholders are bare words to `sh`, so it parses and lints like anything else.

`setup-npu.sh` needed two `# shellcheck disable=SC2016` lines. the backticks in those two printf strings are for the person reading the output, and the single quotes are what keeps them literal, so the warning is asking for the opposite of what the line wants. disabled in the file rather than excluded in the workflow, so the rule still applies to the rest of the script.

left off deliberately: `cross-sysroot.sh` and `systemd-test.sh`. they only ever run on a dev machine or in CI, so the argument for the rest of the list does not reach them. happy to add them if you would rather the list were simply everything.

i ran the step as written against all 17 files and it passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,14 +95,29 @@ jobs:
       #
       # POSIX sh, not bash: the script is piped to `sh` by the documented one-liner, and
       # a bashism would work on a dev box and fail on the board.
+      #
+      # `setup-npu.sh` and `setup-rkaiq.sh` are fetched standalone with curl, exactly like
+      # `setup-gstreamer.sh` two names to the left, and the update hook runs both. `robot-boot-check`
+      # runs on every boot. `hooks/postinstall` runs on every board on every update, which
+      # CONTRIBUTING calls the only thing that does. Each of those is the reason `robot-rescue` is on
+      # this list, and none of them was.
       - name: Lint the installer
         run: |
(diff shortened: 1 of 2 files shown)


Nixxx19:
closing this one. nothing here is actually broken. every script i listed passes shellcheck as it stands, so this is a suggestion about coverage rather than a fix for a defect, and it should not have been filed as an issue. sorry for the noise.

PR #195, opened 2026-09-01
robotd: a mode switch waits for what it needs and never stands a limp robot up

### LINKED ISSUE #194: robot.setMode goes wrong from three states: blind, limp, and shutting down
the mode switch block in `control_loop` (`robotd/src/main.rs`, where `take_mode_switch` is handled) acts on the request in the same tick it arrives, from whatever state the robot happens to be in. three of those states go wrong.

on a tick with no sample, the request is lost for good. the block arms `mode_change` and then only sets `Bringup::Homing` if there is a sample to ramp from. nothing looks at `mode_change` again until the robot next homes for some other reason, and every later `robot.setMode` is refused as "a mode switch is already in flight". so one bus dropout past the coast, during a held d-pad, and the mode cannot be changed again without `robot.init` or a restart.

on a limp robot, it goes `Homing` and then `Ready` with no torque write. `Ready` is the state that means torque on and at home, and it is what the enable driven bring-up checks for, since that only runs from `Limp`. so switch mode first and press start second, which is the natural order for someone who has just put the wheels on, and the policy drives a robot whose servos were never powered. pressing start again finds `Ready` and does nothing. `robot.init` from `Ready` re-homes without touching torque either. the only way out is `robot.relax` first, and nothing says so.

during a shutdown sit it is not refused, so it sets `Homing` and the robot ramps back to the standing pose at gain, torque on, until the sit timer cuts the torque out from under it. that is #159 by a third door. #170 gates `robot.init` and the enable driven bring-up for exactly this, and this path has the same hole.

all three reproduce in the loop's own test harness with `FakeIo` and no policy, except that the sit half of the third one needs the sitstand network, so its test drives the sibling path where `robot.shutdown` powers off at once. the tests are on the branch i am opening against this.

on main at 9f7eaad.

Nixxx19:
fixes #194

three ways a mode switch goes wrong.

if it lands on a tick where the bus read failed it's just lost. nothing picks it up again, so every later switch is refused as already in flight until someone re-homes or restarts.

on a limp robot it goes to Ready without ever turning the motors on. the next Start then finds nothing to bring up and hands the policy a robot that can't move.

during the shutdown sit it stands the robot back up at gain until the sit cuts the torque out from under it.

a test each. also reproduced the limp one on the twin: relax then enable and the duck stands at 0.116 m, same two calls with a setMode in between and it stays on the floor at 0.031 m while enable still says it's driving.


Nixxx19:
merged main in. the conflict was the swap growing a `policy_slots` publish under #191, so `load_mode` now carries `slot_errors` and publishes it the way the ramp end does. nothing else moved.

worth saying after 7563b4c: main now gates `robot.init` and the enable driven bring-up on `powered_off`, and the mode switch is still the door that is not gated. i grafted just the three tests onto main at bc41fb5 and all three fail there for the reasons in their names, so the holes are the same ones on the new code.

`cargo test -p robotd` is 145 green with the merge, clippy with `-D warnings` and fmt clean.

Nixxx19:
merged main in again, the bus trait grew `reboot` and the fake bus in the tests needed it. nothing else changed. `cargo test -p robotd` is 146 green, clippy with `-D warnings` and fmt clean, and grafting just the three tests onto main at 5984efb still fails all three.

one thing worth having on the record before this gets read. #199, opened the morning after this, fixes two of the same three holes in its first commit, the limp one and the blind tick. same diagnosis, different answer, and neither of us saw the other.

where they differ. on a limp robot #199 keeps the switch queued and lets the enable path's ramp finish it, so `robot.mode` reports the old mode until somebody presses start. this swaps the bundle at once and stays limp, so `robot.mode` is true as soon as the request is accepted, and the enable path then does what it always does. on a blind tick #199 refuses and the pad has to hold again, this waits for the next sample. and only this one gates the shutdown sit, which is the door 7563b4c left open after closing the other two.

if you would rather take #199 for the two they share, i will cut this down to the shutdown gate on top of it. your call, either is fine by me.

Nixxx19:
ran this on the twin, now that duck-sim has landed and there is finally a duck here to break.

control, fresh duck: `robot.relax`, wait, `robot.enable {"on": true}`. the journal shows `torque on=true` and the duck stands, trunk 0.116 m, gravity z -1.000.

the same two calls with one `robot.setMode roller` between them: `robot.relax: torque off`, then `mode switch complete`, then `at the home pose; the policy has the robot`, and no `torque on=true` anywhere after the relax. the duck does not stand, trunk 0.031 m, gravity z 0.092. `robot.enable` answered `accepted: true, reason: "enabled — driving"` while the motors were off.

so the home ramp ran into dead servos, the daemon announced it had the robot, and the enable that would have powered the joints was skipped because bringup was no longer `Limp`. that is the limp case this pr and #199's first commit both describe, on a running robot instead of in a test. the order is the natural one for somebody who has just put wheels on a duck: switch the mode, then press start.

to be exact about what that shows: it reproduces the bug, not that this fix is the better of the two. no change pushed here. the offer from my last comment still stands, if you would rather take #199 for the limp and blind-tick cases i will cut this down to the shutdown gate, which is the part only this one has.

Nixxx19:
closing this in favour of #199's first commit, which fixes the limp and blind-tick cases this one does. two fixes for one bug is a decision you should not have to make, and theirs was open first for the two we share.

the third case here, a mode switch during a shutdown sit, is not in #199, so it has moved to #228 with its test. that leaves one pr in this area instead of two that collide.

the reproduction stands whichever lands: on the twin, `robot.relax` then `robot.enable {"on": true}` logs `torque on=true` and the duck stands at trunk 0.116 m, and the same two calls with one `robot.setMode` between them log no torque write at all, reach `at the home pose; the policy has the robot`, and leave the duck at 0.031 m with `robot.enable` answering `accepted: true, reason: "enabled — driving"`.

no hard feelings about the tie, and thanks for duck-sim, it is what turned this from an argument about code into two numbers.

Nixxx19:
reopened, and a correction: i closed this an hour ago saying #199 was open first for the cases we share. that is wrong and i should have checked before writing it. this pr and its issue went up at 21:51 on 1 sep, #199 at 07:12 the next morning. the mistake was mine, not a claim about anyone's work, and #199's fix stands on its own merits either way.

nothing else changed. this is the same three gates it was, and #228 is back to the two it had. the reproduction from earlier in this thread is unaffected.

the offer i keep making still stands and i will stop repeating it after this: if you would rather take #199 for the limp and blind-tick cases, say so and i will cut this down to the shutdown gate, which is the one case #199 does not cover. what i should not have done is make that call for you on a fact i got backwards.

Nixxx19:
hi @pierre-rouanet, mind taking a look at this one when you get a chance? it fixes three ways a mode switch goes wrong, and i reproduced the worst one on the twin.

pierre-rouanet (approved):
Thanks @Nixxx19 and sorry for the review delay!

PR #197, opened 2026-09-01
updater: boot recovery waits for a robot that is still starting, the way the gate does

### LINKED ISSUE #196: boot recovery reverts a release because robotd had not ticked yet
boot recovery can revert a release the robot is about to be healthy on, because it asks `robotd` once and takes "still starting" as the answer.

`robotd` answers "control loop has not completed a cycle yet" from the moment its socket opens until its first tick, as `healthy: false, degraded: false`. the comment on that line in `robotd/src/main.rs` says why that is fine: `"Starting" is not "started". The gate polls, so it will see the transition.` the apply gate does poll, for the configured health timeout at half second intervals, and `updater/src/engine.rs` makes the same three way decision in `recover_on_start` "in the same words, because the reasoning is identical". but it asks once, with `ROBOT_QUERY_TIMEOUT` of two seconds, and anything that is not healthy or degraded falls through to the revert. so at boot a robot that is merely late is a robot that failed.

it is reachable. `robotd.service` is `After=local-fs.target` and `updaterd.service` is `After=network-online.target`, with nothing ordering one against the other, and `robotd` loads its policy networks before the loop ticks, which is seconds on the board. on the boot that exhausts a trial, if the question lands in that window, the release goes back to the previous one with "never reported healthy across 2 boots" in the log, on a robot that would have answered healthy a poll later. the comment above that branch says exactly why that outcome is the one to avoid: an apply killed before its gate ran is the usual reason a trial is never confirmed, it does not mean the release is bad, and reverting one that is working replaces the code under whoever is using the robot, having told them nothing.

it reproduces in the engine's own tests with a robot that answers "starting" once and healthy after. on main the second `recover_on_start` returns `RolledBack { attempted: 1.1.0, reverted_to: 1.0.0, reason: "never reported healthy across 2 boots" }`. the test is on the branch i am opening against this.

on main at 9f7eaad.

Nixxx19:
fixes #196

the apply gate polls robotd for the whole health timeout. boot recovery asks once with a 2s timeout and reverts on whatever it hears, so a robot that's still loading its policies reads as a release that failed.

pulled the gate's polling loop into a helper and had boot recovery ask through it too.

one cost: on a boot where the trial is exhausted and robotd isn't answering at all, recovery now waits the health timeout instead of 2s before reverting. that's the wait the apply gate already does on the same question. board-test is unaffected, its fixture uses probe = none.

### DIFF
--- a/updater/src/engine.rs
+++ b/updater/src/engine.rs
@@ -1706,9 +1706,12 @@ impl Engine {
             //
             // Only for a socket gate. `HealthCheck::None` has nothing to ask, and `Command` answers
             // a two-way question — there is no "degraded" in an exit status.
-            if let HealthCheck::Socket { .. } = cfg.health {
-                match self.robot.health(ROBOT_QUERY_TIMEOUT).await {
-                    crate::robot::Health::Healthy => {
+            if let HealthCheck::Socket { timeout, .. } = &cfg.health {
+                // For as long as the gate would have, and for the gate's reason: at boot this
+                // question lands while `robotd` may still be loading its policies, and the two
+                // units are not ordered. See `robot_verdict`.
+                match self.robot_verdict(*timeout).await {
+                    Some(crate::robot::Health::Healthy) => {
                         tracing::warn!(
                             component = %pending.component,
                             version = %pending.version,
@@ -1719,7 +1722,7 @@ impl Engine {
                         self.boot_counter.confirm(&pending.component)?;
                         continue;
                     }
-                    crate::robot::Health::Degraded(reason) => {
+                    Some(crate::robot::Health::Degraded(reason)) => {
                         tracing::warn!(
                             component = %pending.component,
                             version = %pending.version,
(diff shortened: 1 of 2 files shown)


Nixxx19:
hi @pierre-rouanet, mind a look at this one when you get a chance? it stops the updater rolling back a release that was only slow to start.

pierre-rouanet (approved):
Good catch, and the diagnosis is exactly right. `robotd` says on that line that the gate polls and will see the transition, so anything deciding on that answer has to be the thing that polls — boot recovery was not, and a robot that was merely late read as a release that failed. I reproduced it on `main` before the fix: `RolledBack { attempted: 1.1.0, reverted_to: 1.0.0, reason: "never reported healthy across 2 boots" }`.

One helper shared by both callers is the right shape. And polling on `Unreachable` rather than short-circuiting it is the part that is easy to get wrong: at boot, a socket that is not open yet and a `robotd` that is never coming back are the same answer, and only waiting separates them.

I have pushed two things to the branch.

**Rebased onto `main`.** It was 400-odd commits behind and no longer compiled: `RobotClient` has gained a required `reload_policies`, and `StartingThenHealthy` did not implement it. Folded into your commit so the branch builds at both, rather than fixed in a commit after it. Clean rebase otherwise.

**A doc commit.** The cost is larger than the description prices it at, and it needed writing down rather than only mentioning. `health.timeout` is `30s` on a board, and `recover_on_start` runs *before the socket is served* — so on the boot that reverts with `robotd` genuinely dead, `updaterd` does not answer `/run/updaterd.sock` for 30s, which is exactly when someone is trying to fix the robot. `Type=exec`, so systemd will not kill it and nothing breaks, but `duckctl` hangs. The 2s fixture timeout hides this. I still take the trade — 30s once, against silently replacing a release that was working — but "the wait the apply gate already does" undersells it, because the gate is not holding the recovery socket shut while it waits.

So: the paragraph went in `updater-design.md` §8, which owns the boot counter and still said "ask the robot and then decide"; one sentence in `restart-order.md` §4 step 5, where the consequence belongs because that page is the one stating this runs before the socket is served; and five lines at the call site.

Two smaller notes, neither blocking and neither acted on:

- `robot_verdict` with `timeout = 0` never asks at all and returns `None`. No regression in the gate — same loop shape as before — but boot recovery used to always get one answer, and with `timeout = "0s"` it now reverts without asking. Only reachable from a config that says the gate is zero-length, and "no gate" is a fair reading of that, so I would leave it.
- The new test sleeps a real `HEALTH_POLL_INTERVAL`. `#[tokio::test(start_paused = true)]` would make it instant. The surrounding gate tests sleep for real too, so it is consistent with the file as it stands — take it or leave it.

Green on `check`, `board` and `coverage`. Approving.

PR #198, opened 2026-09-02
updater: three correctness fixes in rollback journaling and apply

hadelan:
Three independent fixes in the updater, one commit each, every one with a regression test.

- **Journal the boot-exhaustion revert and the rescue against the version that failed.** `recover_on_start` wrote its entries by hand instead of through `record()`, and `record_rescue` journalled the rescue against `crumb.to` — the version that *booted* — rather than `crumb.from`, the one whose failure caused it. A later reader of the journal would pair the rescue with the wrong release.
- **Cut hook output on a char boundary.** A bare `truncate(MAX)` on captured hook output panics when the limit lands mid-character. Now it backs off to the boundary.
- **Roll back and journal a select whose apply action fails after the swap.** `transition_to` propagated a post-swap apply-action failure with `?`, leaving the new slot swapped in with the gate unmet — the next start would boot a version whose apply never completed. The failure now joins the same rollback-and-journal path as a failed swap, with a `fail_apply_action` fault-injection switch covering it in tests.

`cargo test --workspace`, `clippy -D warnings` and `fmt --check` are all green; each commit also builds and passes on its own, so the branch bisects cleanly.


pierre-rouanet (approved):
Approving. I checked the three fixes against the code rather than the descriptions, and ran the suite locally.

**The bugs are real.**

`Journal::known_bad` (`updater/src/journal.rs:140`) reads `entry.to` and blacklists the version named there when the newest outcome for it is `RolledBack`, and `Engine::record`'s doc comment already states that a rollback's `to` is the version that *failed*. Both hand-built entries broke that rule, so a boot-exhaustion revert or a rescue blacklisted the release the board had just landed on and left the one that actually failed clean. Per `docs/design/updater-design.md` §8.1 that keeps the failed release eligible for unattended re-apply — the download/rewrite/rollback loop the guard exists to prevent — and it makes `previous_is_usable` skip a perfectly good previous release and escalate to golden.

`String::truncate` panics off a char boundary, and `from_utf8_lossy` pads invalid bytes into 3-byte U+FFFD, so straddling 8192 is easy; the panic unwinds through `Engine::apply`. The backoff loop terminates, 0 always being a boundary.

`transition_to`'s post-swap `run_apply_action` failure returned early after `confirm()` — symlink on the new release, trial disarmed, nothing journalled — while `apply` routes the identical failure into rollback via `post_swap`. Folding it into the same path is right, and `rollback_to` does the disarming, so dropping the explicit `confirm()` is correct.

**Follow-up in 0b898ba.** Fixing `to` left both fields holding the same version on the two paths that build the entry outside `apply`, and `robotctl/src/show.rs:393` renders the pair — a rescue printed `1.1.0 → 1.1.0: ROLLED BACK`, which reads as though nothing moved and drops the release the board landed on. `from` now carries `pending.previous` on the boot-counter revert, which is the line the interrupted apply would have written, and `None` on the rescue, whose crumb holds no such version.

**One non-blocking observation.** Routing `Stuck` through `record` changes its shape: the hand-built entry wrote `to: None`, which `known_bad` skipped entirely, while `journal_outcome` gives `(Some(version), Aborted)`. Since `known_bad` keeps only the newest entry per version and only `RolledBack` counts as bad, an `Aborted` entry now masks an earlier `RolledBack` for the same version and un-blacklists it. The `apply` and `transition_to` paths already behave that way, so this is consistency rather than a new bug — but "an abort clears a prior rollback verdict" is arguable on its own merits, and worth a separate look.

Locally: `cargo test -p updater` green (224/12/80/7/12/26), `clippy -p updater --all-targets -D warnings` and `cargo fmt --check` clean.

PR #199, opened 2026-09-02
robotd, kinematics, robotd-params: four hardening fixes for the drive loop and startup

hadelan:
Four independent fixes, one commit each, every one with a regression test.

- **Never start the homing ramp for a mode switch while limp.** A mode switch queued with torque off jumped straight to `Homing`, skipping the `set_torque` the enable path owns — the ramp "finished" over dead motors, the other mode's policies loaded, and the state reported `Ready`/`homed` for a robot lying on the floor. The same request on a tick with no position sample wedged `mode_change` forever, refusing every later switch as already in flight. `Limp` now stays `Limp` (the enable path turns the motors on and ramps; the queued switch completes when that ramp does), and a sample-less tick refuses the switch outright.
- **Drop non-finite command targets instead of poisoning the command filters.** JSON parses `1e400` as infinity, so a client can put one on the wire. Folded into the twist/head/body EMAs it is permanent (`ema += α·(inf − ema)` is `inf` on every later tick), and since the safety layer refuses non-finite joint targets rather than clamping them, a single bad `robot.move` froze the robot on its hold pose until reboot. The filters now drop non-finite targets.
- **An empty band is no hand even when `min_zones` is 0.** `min_zones = 0` is a legal config value, and it made the length check pass on an empty band — straight into the low-percentile index, which panics on an empty `Vec`. The check is now floored at one zone.
- **Refuse `policy.walk = "none"` at load instead of panicking.** The `"none"` sentinel legitimately disables the optional policy slots, but `walk` is the one slot `resolved()` cannot leave empty, so the config got past load and panicked at startup. `validate()` refuses it now, as a config error with the path attached.

`cargo test --workspace` (1035 passed, 0 failed), `clippy -D warnings` and `fmt --check` are all green.


dufudan:
These fixes highlight an important portability question: the policy is only one part of the runtime contract. I’m evaluating a Linux + RTOS SoC architecture and would be interested in which robotd safety and state-machine responsibilities you would keep beside the policy loop, versus move to a real-time core.

hadelan:
Hi, @pierre-rouanet  a friendly ping on this PR. It bundles four independent hardening fixes (one commit each, each with a regression test) — happy to split them into PRs if that's easier to review. Thanks!

pierre-rouanet:
Hi @hadelan ! And thanks for the contribution! 
Yes it would be better to split into different PRs. And btw the fix 4 was already done in 7d8296e

hadelan:
Hi @pierre-rouanet , thanks for the quick reply! I've split the PR as suggested:

- #237 — homing ramp for mode switch while limp
- #238 — non-finite command targets
- #239 — empty band with min_zones = 0

And dropped fix 4 since 7d8296e already covers it. Closing this one in favor of the three above.

PR #202, opened 2026-09-02
maploc on the MuJoCo twin: a mirrored sensor, a fall at enable, and a route steered on stale odometry

peterschade:
## What this is

maploc run on the MuJoCo twin, with the map drawn into the viewer and scored against the scene's own walls. It stacks on #127 (`maploc`) and on the `sim-remote-io` branch, both by @apirrone, so the diff against `main` carries both; the commits that belong to this PR are the ones after `merge: maploc onto sim-remote-io (investigation)`.

The scaffolding under `sim-maploc/` and `try-maploc.sh` hardcodes paths to a local `microduck_rl` checkout. It is investigation tooling, kept so the runs are reproducible, not something to ship on a robot.

## What the twin found

Three things stood between maploc and a usable map, none of them in maploc.

**The simulated ToF numbered its columns backwards.** `microduck_rl`'s `sim/tof.py` put column 0 on the sensor's right; `kinematics::tof` reads the real sensor's buffer with column 0 on the left. Every frame reached the mapper mirrored: walls seen at an angle were inked at their mirror image across the head's axis, phantom walls stood in open floor, and no loop ever closed on the twin. Fix in pollen-robotics/microduck_rl#30. `examples/evaluate` gained `MAPLOC_MIRROR_COLS=1` so recordings made before the fix still replay correctly.

**The duck fell at enable.** `try-maploc.sh` booted from the STAND keyframe; `robot.enable` then ramps to the home pose with nothing balancing and the duck tips over, as `scripts/duck-sim` already documents. maploc saw the fall, went suspect, confirmed none of its relocalize candidates, and resumed a minute later on raw odometry. Boot seated instead: robotd rises via the sitstand policy.

**The route steered on stale odometry.** `scan_walk.py` read one line per call from the `robot.subscribe` stream, which writes faster than the loop reads; by the second stop the pose it steered on was the stop before. Drain the backlog.

## Numbers

Map wall cells against the scene's walls and furniture footprints (`sim-maploc/apartment.toml`, generated from `apartment.xml`; boot pose exact, no tape measure):

| recording | boot | sensor | mean | p50 | p90 | wall cells on no real wall | loops |
|---|---|---|---|---|---|---|---|
| 1788348816 | STAND, fell | mirrored | 0.141 m | 0.094 | 0.315 | | 0 |
| 1788356722 | seated | mirrored | 0.174 m | 0.114 | 0.386 | 18 % | 0 |
| 1788356722 replayed `MAPLOC_MIRROR_COLS=1` | | corrected | 0.058 m | 0.036 | 0.131 | | 3 |
| 1788358851 | seated | fixed | **0.031 m** | 0.026 | 0.066 | 0 % | 0 |

On the clean run the tracked pose stays within about 6 cm of MuJoCo's ground truth at every stop, including after the return leg, while raw odometry drifts by up to 0.35 m. Before the fix the twin's best was worse than maploc's own hardware figure of 0.126 m; after it, the twin is the first place maploc's mapping has been shown to work.

The three recordings are committed under `recordings/` (about 4.5 MB) so the bench can be rerun without a simulator:

```sh
cargo build --release -p maploc --example evaluate
./target/release/examples/evaluate recordings/1788358851.mdlg sim-maploc/apartment.toml out/
```

## Open

- The bench's return-to-start and kidnap metrics want a sit as the protocol mark; `scan_walk.py` stands at the end, so both report "protocol not detected".
- No loop closed live on the clean run even though the return leg re-sees the boot area; the mirrored replay of the earlier session closed three. Worth chasing offline on `1788358851.mdlg`.
- `scan_walk.py` gates its walk bursts on wall-clock modulo, and a burst shorter than the policy's start latency moves nothing; the clean run stood still for three legs because of it.


apirrone:
Great, i'll test all this when I have time ! Thanks !

PR #207, opened 2026-09-02
btd, mediad: a restarted daemon gets the next request, not the one after

### LINKED ISSUE #206: the first request after a daemon restarts is answered with a broken pipe, over ble and webrtc
both `btd/src/upstream.rs` and `mediad/src/upstream.rs` keep one connection per service and lane for the length of a client session. when the daemon on the other end restarts, the pool's reader task sees the socket close and exits, but the entry stays in the pool. the next request on that lane is written into the closed socket, fails with a broken pipe, and only then is the entry dropped. the comment on that path says a broken pipe is ordinary, the daemon restarted, and the next call reconnects. it is the call after the next. the one that hit the broken pipe is answered with `INTERNAL_ERROR`, "Updater is not answering: Broken pipe", to a phone or a console, about a daemon that is up and listening on a fresh socket.

it is deterministic, and it sits on the path ble exists for. `update.apply` from the phone: the release's postinstall hook restarts `updaterd` mid apply, which the engine's own comments call the usual reason a trial is never confirmed. the phone's next `update.status` on that lane fails. every lane to that service fails its next request the same way, so it is one error per lane, not one. `robotd` is the daemon an update restarts, so the console's next `robot.health` after any update fails the same way over webrtc.

the bytes that hit the broken pipe never reached anything, so there is nothing unsafe about sending them again on a fresh connection, and a daemon that is genuinely gone fails that reconnect with the error that is actually true.

both copies reproduce with a fake daemon that answers once and hangs up, then a fresh listener at the same path. on main the second `send` returns `Os { code: 32, kind: BrokenPipe }`. the tests are on the branch i am opening against this.

on main at 2c61dcc.

Nixxx19:
fixes #206

a write that fails because the peer went away is written once more on a fresh connection, in both pools. those bytes never left, so there is nothing to double up. a timeout is not retried, that is a daemon that is there and stuck, and a daemon that is genuinely gone fails the reconnect with the error that is actually true. the write itself moved into one bounded helper that drops the connection on any failure, which is what the old code did inline.

the same change in both files rather than one shared crate, because the two pools are already two copies and unifying them is a different change.

two tests, one per copy, each with a fake daemon that answers once and hangs up, then a fresh listener at the same path. both fail on main with `BrokenPipe` on the second send and pass now with the request reaching the new listener and its answer forwarded.

`cargo test -p btd`, `cargo test -p mediad`, clippy with `-D warnings` and fmt are green.

### DIFF
--- a/btd/src/upstream.rs
+++ b/btd/src/upstream.rs
@@ -153,30 +153,47 @@ impl Pool {
     }
 
     /// Send one line to `upstream` on `lane`'s connection, connecting first if needed.
+    ///
+    /// A daemon that restarted between two requests is ordinary here: `robotd` is the one an
+    /// update restarts, and `updaterd` restarts itself from a release's postinstall hook. The
+    /// connection from before the restart is still in the pool, its reader has already seen the
+    /// socket close, and the first write into it fails. Those bytes never left, so they are written
+    /// again on a fresh connection rather than reported. Reporting them told a phone that a daemon
+    /// listening on a fresh socket was not answering, once per lane, after every update. Once and
+    /// not in a loop: a daemon that is genuinely gone fails the reconnect, and that is the error
+    /// worth reporting.
     pub async fn send(&mut self, upstream: Upstream, lane: Lane, line: &str) -> io::Result<()> {
         let key = (upstream, lane);
+        let mut bytes = line.as_bytes().to_vec();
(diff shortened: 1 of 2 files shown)


pierre-rouanet (approved):
Thanks @Nixxx19

PR #208, opened 2026-09-03
padd: refuse --hz 0 instead of dividing by it

### LINKED ISSUE #184: padd --hz 0 panics at startup instead of refusing the flag

pierre-rouanet:
fixes #184

`--hz` was a plain `u32`, so `--hz 0` reached `Duration::from_secs_f64(1.0 / 0.0)` and panicked once the daemon had already brought up the pad and connected to robotd.

the flag is now bounded at both ends, 1–1000, and refused rather than clamped: a rate quietly changed to something else is one nobody can explain later. the ceiling is the other half of the same line — a rate this loop cannot keep gives a period of 0 ns, `checked_sub` never has anything left to sleep on, and the pad spins on robotd's socket. that is the range `control.hz` already rejects outside of.

the test parses the flag at 0, 1, 1000, 1001, the top of a `u32`, and with no flag at all.

`cargo test -p padd` is green, as are `cargo fmt --check` and `cargo clippy --all-targets`.

carries @Nixxx19's commits from #187 unchanged, on top of main — #204 rewrote the `--hz` doc comment, so the bound and the doc needed reconciling and the PR branch was not pushable from here.

### DIFF
--- a/padd/src/main.rs
+++ b/padd/src/main.rs
@@ -111,10 +111,13 @@ struct Args {
     /// How often to send intents. Matching the control rate exactly buys nothing — the loop
     /// reads the latest value once per tick — but staying at or above it keeps the added
     /// latency under one tick.
-    ///
-    /// refused below 1 rather than clamped: zero reaches `1.0 / 0.0` below, and
-    /// `Duration::from_secs_f64` panics on infinity instead of producing a very long period.
-    #[arg(long, default_value_t = 50, value_parser = clap::value_parser!(u32).range(1..))]
+    // Bounded both ways, and refused rather than clamped so the flag says what it did.
+    // Zero reaches `1.0 / 0.0` and `Duration::from_secs_f64` panics on infinity. The ceiling
+    // is the other half of the same line: a rate this loop cannot keep gives a period of 0 ns,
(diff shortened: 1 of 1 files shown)


pierre-rouanet:
Thanks again @Nixxx19

PR #211, opened 2026-09-03
robot.soften: let go softly, on the pad's Select

RemiFabre:
`robot.relax` cuts torque at once and the robot drops. This adds the gentle version, for a robot forcing against something: you want it to let go *now*, but not to fall like a stone.

## What

- `robot.soften`: gain to `gain_limp` (50) at once, then down to zero over 1 s (in steps of 5, so it is not a register write on every tick) while the joints are commanded where they are, then torque off. Same end state as `relax`: the robot is limp, `init` or Start brings it up. Refused over Bluetooth like `relax`, allowed on the console like `relax`.
- **Pad: Select** = soften. Select held 2 s still powers off (the motors are already let go by then).
- `sudo robotctl robot soften` for the bench.

## How to test
Hold the robot with the policy on, press Select: it goes soft within a second and the joints are free. Press Start to stand up again. Select held 2 s: powers off as before.

### DIFF
--- a/btd/src/route.rs
+++ b/btd/src/route.rs
@@ -354,7 +354,7 @@ fn permits(call: &proto::Call) -> bool {
         // offer, and `robot.init` is its counterpart: standing a robot up moves every joint at once,
         // which wants the person doing it to be looking at the robot rather than at a screen. Both
         // are `robotctl` on the robot, deliberately.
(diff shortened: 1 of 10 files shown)


RemiFabre:
Dropped: a plain robot.relax on Select is enough. Replaced by a smaller PR.

PR #216, opened 2026-09-06
duck-control: the safety layer says what it refused, clamped or zeroedfeat(safety): say what the safety layer refused, clamped or zeroed

yunloong-Y:
The safety layer owns the only write handle to the motors, and it silently rewrote or
refused commands on three paths: a non-finite target was replaced by the hold pose,
out-of-range targets were clamped to the actuator's travel, and a stale intent had its
velocity zeroed.
Each of those recorded a Limit in Applied — which reaches a client only while one is
subscribed to robot.state, and on a robot nobody usually is. So a robot that would not do
what it was told left no trace anywhere, and "why is it ignoring me" had nothing to go on.
Each now says so, rate-limited the way the bus-read warning already is: the first trip, then
every 50th — at 50 Hz, one line a second. A limit that fires once is news and must appear;
one that fires every tick is a state, and fifty identical lines a second is what teaches
everyone to stop reading the journal.
The fall verdict is logged on every flip with no rate limit, because it is already debounced:
a flip is the rare event, and it is the only record of a fall there is.
Three details worth calling out:
- Runs, not totals. Each limit's counter is cleared by a tick that does not trip it, so
the number reads "how long has this been going on" rather than "how many times since
Tuesday" — the same reason the fall debounce resets.
- Three counters, not one. A deadman that fires for a second says nothing about whether a
joint is being clamped. One shared counter would let whichever fires most starve the other
two of a line entirely.
- Clamps report how many and which. One joint at its limit and nine are different faults
wearing the same Limit::Range on the wire, and the lowest index out is what someone goes
to look up.
gate takes &mut self because how long intents have been stale is now its to remember.
No behaviour changes: what is written to the bus is exactly what it was before, which two of
the new tests exist to say.
Verification
- cargo test -p duck-control -p robotd — 215 passed, 0 failed
- cargo fmt --all -- --check — clean
- RUSTFLAGS="-D warnings" cargo clippy -p duck-control -p robotd --all-targets — no warnings
Five tests added, covering the rate limit itself (1, then every 50, nothing in between),
a run resetting when the limit stops firing, the three counters staying independent, a plain
tick tripping and reporting nothing, and the two applied outputs — refused targets writing
the hold pose, clamped targets landing inside the actuator's travel.

### DIFF
--- a/duck-control/src/safety.rs
+++ b/duck-control/src/safety.rs
@@ -102,6 +102,49 @@ impl Applied {
     }
 }
 
+/// How often a limit that keeps firing is said again.
+///
+/// The first tick a limit trips always reports — a clamp that happens once is news, and
+/// before this none of it was visible at all. After that, one line per fifty: at 50 Hz that
+/// is one a second, which shows a fault is still going on without writing a line per tick
+/// into the journal. The same shape as the bus-read warning in `robotd`, which reports the
+/// first failure and then every tenth.
+const LIMIT_LOG_EVERY: u64 = 50;
+
+/// How many ticks running each limit has tripped, which is what the journal is keyed off.
+///
+/// Three counters rather than one because the three limits are independent events: a
+/// deadman that fires for a second says nothing about whether a joint is being clamped, and
+/// one shared counter would let whichever fires most starve the other two of a line.
+///
+/// Each is cleared by a tick that reaches that limit's check and does *not* trip it, so the
+/// number reads "how long has this been going on" — which is the number worth saying, and
+/// the one a reader has to be able to distinguish from "how many times has it ever
+/// happened".
+///
+/// A tick refused as non-finite returns before the clamp is ever reached, so it leaves
+/// `range` standing rather than clearing it. That is deliberate, and the alternative is
+/// worse: a target alternating between `NaN` and out-of-range would end the range run on
+/// every second tick, and both limits would then report on every tick they tripped — the
+/// line per tick this counting exists to prevent. So the number is trips in a run, not
+/// strictly consecutive ticks.
+#[derive(Debug, Default, PartialEq)]
+struct LimitRuns {
+    deadman: u64,
+    range: u64,
+    not_finite: u64,
+}
+
+impl LimitRuns {
+    /// Whether a run of `n` has reached a point worth writing down.
+    ///
+    /// Kept here rather than at each call site so the three limits cannot drift into
+    /// three different ideas of "often enough".
+    fn worth_logging(n: u64) -> bool {
+        n == 1 || n.is_multiple_of(LIMIT_LOG_EVERY)
+    }
+}
+
 pub struct Safety<T: RobotIo> {
     io: T,
     config: SafetyConfig,
@@ -111,6 +154,12 @@ pub struct Safety<T: RobotIo> {
     /// Tracks the last gain written so an unchanged one is not rewritten every tick — that
     /// would be fifteen bus writes per tick for no reason.
     gain: Option<u16>,
+    /// Trip counts, for rate-limiting what the layer says about itself.
+    runs: LimitRuns,
+    /// Whether an intent has ever arrived inside the deadman, which is what makes a stale
+    /// one news. Until one has, the robot has no driver to have lost. See
+    /// [`Self::deadman_is_news`].
+    deadman_armed: bool,
 }
 
 impl<T: RobotIo> Safety<T> {
@@ -121,6 +170,8 @@ impl<T: RobotIo> Safety<T> {
             falling_for: Duration::ZERO,
             fallen: false,
             gain: None,
+            runs: LimitRuns::default(),
+            deadman_armed: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


pierre-rouanet (changes requested):
The fall-verdict line I'd take as-is. `fallen` reaches the journal only as a sampled field in the five-minute loop summary (`robotd/src/main.rs:2903`), so a fall inside one window left no record and one that persisted had no time on it. Logging the flip closes that, and it is debounced enough not to need a rate limit.

The rest needs another look.

**The deadman line will fire once a second, forever, on an idle robot.** `Intents::new` stamps the twist at `at_us: 0` on purpose — `robotd/src/intents.rs:258`, "before any client connects the twist already reads as maximally stale" — and `gate` is called unconditionally every tick at `main.rs:1826`, enabled or not, torque or not. So from ~500 ms after boot:

```
WARN intents went stale — the velocity command is zeroed intent_age_ms=412037 deadman_ms=500 consecutive=20601
```

~86k lines a day, `consecutive` climbing without bound, and the run counter never resetting because nothing ever sends a twist. That is a bench robot, and any robot driven through skills or `robotctl` rather than a pad. `padd/src/main.rs:733` says the same from the other side: its 100 ms heartbeat exists so a *connected* idle pad does not read as deadman-limited — with no pad, it always does. And `main.rs:66` is already the argument against this: per-tick logging "is what *evicts* the logs support needs". 1 Hz is on the same axis, just slower.

Suggest not reporting a deadman that has never been fresh: a robot with no driver is not being ignored, and that is not news. Worth a test that a never-driven robot logs nothing.

**"'why is it ignoring me' had nothing to go on" is not quite right**, and it is the justification for the other three lines. `robotctl monitor` subscribes to `robot.state` and renders the limits as an always-present row, explained in prose rather than as tokens — `robotctl/src/monitor.rs:1731`, `explain_limit` at `monitor.rs:2737`:

```
 limits  none — the command went through untouched
 limits  deadman — no intent arrived recently, velocity zeroed
```

The live case is covered, and the premise is weakest where it matters: the moment someone wants to know why the robot is ignoring them is the moment they attach a client. What is genuinely missing is the *post-hoc* case — nobody was watching and you want to know afterwards. That is a narrower claim but a real one for the non-finite and range lines: a policy emitting `NaN`, or driving past ±π, should never happen and is worth evidence of after the fact. Worth restating the body around that rather than around the live question.

Smaller:

- The fall debounce is one-directional — 200 ms of down to latch, but a single upright sample clears it (`safety.rs:247`). A robot held near the threshold with IMU noise can cycle every ~220 ms, so a flip is not unconditionally rare, and the comment at `safety.rs:252` reads as if it were.
- A non-finite tick does not clear the range run: `apply` returns at `safety.rs:334` before `self.runs.range = 0`. `LimitRuns`' doc says each counter is cleared by a tick that does not trip that limit, and a NaN tick does not trip `Range`. The effect is cosmetic — 49 clamped ticks, a NaN tick, one clamped tick logs as `consecutive=50` — but `the_three_limits_are_counted_separately` pins the frozen behaviour. Either reset before the early return, or narrow the doc to say a run ends only on a tick that reaches the clamp.
- The title is two titles concatenated.

Tests and `fmt` are clean locally, 215 passed. The five tests are well aimed, particularly the two asserting the bus writes are unchanged — that is the claim that matters most for a diff inside this layer.

yunloong-Y:
Thank you — the deadman point is right and I had not seen it. I checked all three
citations: `Intents::new` does stamp the twist at `at_us: 0` deliberately, `gate` is called
unconditionally at main.rs:1826, and the comment above `LOOP_SUMMARY_INTERVAL` does already
make the argument. 1 Hz forever on an idle robot is that same failure at a slower speed.

Fixed as you suggest: `deadman_armed` is set by the first fresh intent and the line is
gated on it, so a robot that has never had a driver says nothing. New test
`a_robot_that_has_never_been_driven_reports_no_deadman` — 200 stale ticks on a fresh
`Safety`, nothing is news; one fresh intent arms it and going quiet is news again.

On the framing, agreed, and I had it wrong. `monitor` renders the limits as prose whenever
someone is attached, so the live case is covered and "why is it ignoring me" was the wrong
justification for these lines. Body restated around the post-hoc case: nobody was watching,
and afterwards you want to know. That is the claim the non-finite and range lines actually
support, and the one I should have made.

The three smaller ones:

- **Fall debounce is one-directional** — correct, and the comment did claim a flip was rare.
  Rewritten to say a robot held near the threshold can cycle in roughly the debounce period,
  and why it is still not rate-limited: the line reports a change, not a state, so dropping
  all but one in fifty would drop the one that says *when*. Left unlimited per your call —
  say the word and I will rate-limit it.
- **Range run** — took the second option. Resetting before the early return is worse in the
  case that matters: a target alternating between `NaN` and out-of-range would end the run
  every other tick, and then both limits report on the first tick of every run — a line per
  tick, which is what the counting is there to prevent. So the behaviour stays and the doc
  now says what it is: a run ends on a tick that reaches the check, and the count is trips
  in a run rather than strictly consecutive ticks. Log field renamed `consecutive` →
  `tripped` to match. New test `a_refused_tick_does_not_end_a_range_run` pins it.
- **Title** — now `duck-control: what the safety layer refuses or rewrites never reaches the
  journal`.

217 tests, fmt and clippy clean.

pierre-rouanet (approved):
Thanks @yunloong-Y !

PR #218, opened 2026-09-07
robotd: prevent duplicate startup from replacing the live IPC socket

tianrking:
Starting a second `robotd` with the same `--socket` currently unlinks the first daemon's listener after starting another control thread. Both processes stay alive while new clients reach the second process. The duplicate also overwrites the running daemon's PID and build in `/run/robotd/identity.json`, which `robotctl health` and updater startup checks trust. Standalone `robotd init` is another unguarded bus writer: its requirement to stop the daemon first was only documented.

Acquire an advisory lock on `<socket>.lock` before publishing identity or opening the motor bus. The daemon binds its IPC listener before starting the control thread and holds the lock through shutdown and socket cleanup. `robotd init` uses the same lock-only claim and holds it through the entire initialization ramp, so a daemon and an init cannot join each other mid-operation. A refused contender exits with a clear error before publishing identity or touching the bus.

The lock file stays in place so all contenders use the same inode; file existence does not mean the lock is held. Closing the file or exiting releases ownership, including after `SIGKILL`. The design documentation now records this lifecycle and why cleanup must never unlink the lock file.

For an existing socket without a lock, the daemon's bind path uses a bounded connection probe. It removes only an actual socket returning `ConnectionRefused`, preserving live listeners, non-socket paths and ambiguous failures. Standalone `init` only locks: it neither binds nor removes the socket. The advisory lock is per `--socket`; callers using the same physical bus must use the same setting. Older binaries and other tools do not participate and must still be stopped before standalone init. Independent fake daemons can use different endpoints. No dependency or wire-protocol changes.

Validation on Linux x86_64 (Ubuntu 24.04 under WSL):

- The duplicate-daemon regression fails on the original `bc41fb5` base. The three init-related regression cases fail on the previous PR head `e72a489` and pass with the revision.
- `cargo test --locked -p robotd`: **159 passed** (142 unit tests, 10 process-level startup tests, 7 updater-gate tests).
- Startup coverage includes duplicate and simultaneous daemon startup (ten rounds), locks occupied before a socket exists, live legacy listeners including a saturated Linux accept queue, `SIGKILL` recovery, clean restart, non-socket preservation, init refusal against a running daemon, and lock release after init failure. Refusals also check that startup identity was not published.
- `cargo fmt --all --check` and `cargo clippy --locked -p robotd --all-targets -- -D warnings` pass.
- An additional local probe, separate from the Cargo suite, runs the actual `robotd init` against a private pseudo-terminal responding to Dynamixel v2 traffic. It completes a 1.5-second / 75-write ramp, observes the lock at every interpolation write, rejects both a concurrent daemon and another init, preserves the owner's identity, and starts a healthy fake daemon after init returns. A subsequent refused init sends no bus packets. The same probe fails on `e72a489`.

The PTY peer validates process exclusion and lock lifetime, not physical actuator behavior. No physical robot was used. Full-workspace and aarch64/userland checks run in repository CI; see Checks for the current head.

### DIFF
--- a/docs/design/robotd-design.md
+++ b/docs/design/robotd-design.md
@@ -43,24 +43,41 @@ the hardware does: the v2 board sits on the Dynamixel bus and serves an on-chip 
 quaternion out of the same register block the servos answer at. One board, one code path, no
 IMU abstraction. It is listed first in the id vector so it answers before the servo burst.
 
-**One owner at a time, and nothing hard-enforces it.** `serialport` sets `TIOCEXCL`, which
+**One owner at a time; tty exclusivity alone does not enforce it.** `serialport` sets `TIOCEXCL`, which
 turns a second *unprivileged* open into `EBUSY` — but `robotd.service` runs as root, because
-motor control needs the character devices, and root is not stopped by that flag. So the
-exclusion is arranged rather than enforced, and each other claimant is kept off the port
-deliberately:
+motor control needs the character devices, and root is not stopped by that flag. The daemon
+and standalone `init` therefore share an advisory lock, and other claimants are kept off the
+port separately:
 
 - **The control loop** owns it for as long as the daemon runs.
-- **`robotd init`** opens the port itself, and as root it will succeed *while the daemon is
-  running* — two writers interleaving packets on one bus, which reads as a hardware fault. So
-  it wants the daemon stopped, and that is exactly why `robot.init` and `robot.relax` exist as
-  IPC methods (§3.3): the daemon serves both from inside the loop, so nothing else has to open
-  the bus at all. `init` is the escape hatch for a robot whose daemon is not running.
+- **`robotd init`** opens the port itself, but must first take the daemon's endpoint lock.
+  It holds that lock through the entire ramp: a running daemon refuses `init`, and an `init`
+  already moving the robot refuses a daemon startup or another `init`. `robot.init` and
+  `robot.relax` remain the IPC methods (§3.3) for a running daemon; standalone `init` is the
+  escape hatch for a robot whose daemon is not running.
 - **`serial-getty@ttyS2`** — Armbian runs a login console on UART2 by default, and an `agetty`
   holding the port makes every servo invisible to everything else. `scripts/setup-board.sh`
   masks the unit; `fuser -v /dev/ttyS2` naming `agetty` is how that was found, and it is still
   the command that answers "who has the bus".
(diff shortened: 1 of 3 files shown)


pierre-rouanet (commented):
Thanks for this — the diagnosis is right, and the fix is in the right place: claiming the
endpoint *before* `spawn_control_thread` is what matters, because the old ordering meant the
loser had already started a control loop by the time it unlinked the winner's socket. On a real
board that is two processes writing the same Dynamixel bus, and a tty is not exclusive-open.

The reordering of `log_startup_identity!` fixes more than the description claims, and that is
worth saying in the commit message rather than leaving as "before publishing an identity".
`publish_identity` writes `/run/robotd/identity.json` keyed on the service name alone, so under
the old ordering a duplicate overwrote the running daemon's pid and build before it ever touched
the socket — `robotctl health` and updaterd's startup check would both have believed the loser.

One change before this goes in.

## `robotd init` should take the lock too

The PR leaves it with "its existing requirement to stop the daemon first", but that requirement
is a sentence in a comment (`robotd/src/main.rs:4191`) and nothing enforces it. `run_init` opens
the bus itself with `DynamixelIo::open`, enables torque, sets the gain and ramps to the home
pose. Run against a live daemon that is two writers on one serial port moving the same robot —
the exact failure this PR exists to prevent, reached by a shorter path than starting a second
daemon, and the path a person is more likely to take by hand on a bench.

The lock makes that unrepresentable for about three lines. Split the claim in two:

- a lock-only step: open `<socket>.lock`, `try_lock` it, return the `File`
- `claim_socket`: that step, plus the bind

`main` takes the lock before `run_init` and holds it across the ramp; the daemon path calls
`claim_socket` as it does now. A running daemon then makes `robotd init` exit with the same clear
error instead of fighting it on the bus — and because the lock is held for the whole ramp, a
`systemctl start robotd` landing mid-ramp fails cleanly rather than joining in, which is the
better half of the property.

## Smaller

`/run/robotd.sock.lock` is a new persistent on-disk artifact with a non-obvious rule attached:
never unlink it, even at shutdown, or contenders split across two inodes. Right now that rule
exists only in a comment inside `claim_socket`. `docs/design/robotd-design.md` has nothing on
socket ownership at all, and a paragraph there is what stops the next cleanup routine from
deleting the file.

CI is running now.

pierre-rouanet (changes requested):
See comment above.

pierre-rouanet (approved):
See comment above.

pierre-rouanet:
Thanks @tianrking !

PR #223, opened 2026-09-07
A demo Space that processes the duck's camera, and what it measured

pierre-rouanet:
Everything after #219 merged. A Gradio Space on Hugging Face hardware consumes the robot's camera through the rendezvous and runs OpenCV over each frame — and on its first run it measured the thing nothing else could.

## The measurement

```
the rendezvous welcomed this consumer as 9d3a17dd (PierreRouanet)
found olducky (8a73eda9), asking for a session
session 4c7e9c60 started; waiting for the robot's offer
peer connection connecting (ice checking)
the session ended: the robot ended the session      ← 8s later
peer connection failed (ice failed)
```

Signalling crosses perfectly and **no candidate pair works**. Three data points now instead of one: a browser on the robot's LAN connects on host candidates, a 4G phone fails and CGNAT was the suspicion, and **a Hugging Face Space fails too** — and a Space is not behind CGNAT. So TURN is not a fallback for awkward networks; it is the requirement for every consumer that is not on the robot's own wifi, which is every cloud backend and therefore anything doing perception off-robot. §6 carries the log.

Nothing in the demo needs changing for it: **it will connect the day the robot can offer a relay candidate.**

## The consumer is ours

`reachy_mini[central-consumer]` proved the transport — it was the first non-browser client to pull frames out of a duck — and was the wrong *dependency*, three times over. `reachy_mini/__init__.py` imports the daemon's world, so a consumer drags in **PyGObject**: a source build wanting a C compiler and cairo headers for GStreamer bindings it never touches. It pins `starlette<1.0.0`, which no current Gradio can satisfy (`ResolutionImpossible`, correctly). And it ignores a data channel labelled `control`, so the camera geometry cannot be asked for.

`spaces/vision-demo/consumer.py` is 300 lines: SSE in, `POST /send` out, aiortc answering the robot's offer, frames as numpy. Verified against `olducky`:

```
frame 8: (720, 1280, 3) uint8
media.video: {'width': 1280, 'height': 720, 'rotate': 90,
              'intrinsics': {'fx': 1809.52, 'cx': 640.0, 'calibrated': False}}
pc_state: connected  ice_state: completed  control: True
```

That second line is the payoff: the intrinsics come from the robot, so the demo's geometry overlay uses published numbers rather than repeating the arithmetic. The fifteen lines that make DTLS complete against GStreamer's `webrtcsink` are copied from their consumer with the credit and the upstream link (aiortc PR #1392) — they are the reason this took an afternoon rather than a week.

Six dependencies instead of 117: `aiortc`, `av`, `httpx`, `gradio[oauth]`, `opencv-python-headless`, `numpy`.

## What the Space is

`spaces/vision-demo`, deployed by `scripts/publish-space.sh` — source in-repo for §5's reason, since a consumer tracks the rendezvous protocol, this project's method names and the camera's geometry. Docker rather than `sdk: gradio`, because a Gradio Space cannot choose its Gradio version and 6.26 is the one version that cannot work here.

Five filters — raw, Canny edges, motion diff, sparse Lucas–Kanade flow, and an intrinsics overlay — with the pixels in `filters.py` importing only OpenCV and numpy, so every one is exercised on a synthetic sideways frame pair without a WebRTC stack, a robot or a token.

**It authenticates with an `HF_TOKEN` secret and must stay private.** A visitor's own OAuth token would be better and would make it publishable; two attempts failed on platform behaviour rather than code — a static Space never injected the client id (§5.0), and a Docker Space does not put `OAUTH_CLIENT_ID` in the environment either, so Gradio mocks OAuth and refuses to start. A demo that will not start is worse than a demo with one credential.

## Four bugs of mine, three of them the same bug

A hand-kept list that has to match reality, failing silently in the direction of looking fine: `cp "$SOURCE"/*` published whatever was lying around, `git diff --quiet` reported "already serves this" for a file that was only ever added, and `COPY app.py filters.py` built a perfect image without `consumer.py` in it. All three now ask git or the filesystem what is actually there.

The fourth: `stop()` posted its farewell *after* cancelling the event stream, which the service refuses with a 400. I had diagnosed that exact bug in their consumer an hour earlier and said ours could not have it — true of the Rust relay, not of the Python I had yet to write. The ordering is a docstring now.

### DIFF
--- a/docs/design/remote-access-design.md
+++ b/docs/design/remote-access-design.md
@@ -721,6 +721,30 @@ blocks (a `try_read` that yields nothing rather than waiting) and never fails. A
 the ordinary state for the first few seconds after boot and forever on a robot with no account,
 and it means host and srflx only, which is all anything on the same network needs.
 
+**Measured: a data centre cannot reach a robot without a relay.** The demo Space
+(`spaces/vision-demo`) got as far as it can and the log says where that is:
+
+```
+the rendezvous welcomed this consumer as 9d3a17dd (PierreRouanet)
+found olducky (8a73eda9), asking for a session
+session 4c7e9c60 started; waiting for the robot's offer
+peer connection connecting (ice checking)
+the session ended: the robot ended the session       ← 8s later
+peer connection failed (ice failed)
+```
+
+Signalling crosses perfectly — welcome, listing, session, offer, answer — and **no candidate pair
+works**. The robot offers host and srflx; a Space offers its own; neither offers a `relay`. Eight
+seconds in, `webrtcsink` times out a consumer whose ICE never connected and ends the session, which
+the bridge forwards, which is the `the robot ended the session` line arriving before the ICE
+verdict.
+
+That settles what was previously a guess. On one LAN the console works on host candidates; from a
+4G phone it failed and carrier-grade NAT was the suspicion; **from a Hugging Face Space it fails
+too**, and a Space is not behind CGNAT. So TURN is not a fallback for awkward networks — it is the
+requirement for every consumer that is not on the robot's own wifi, which includes every cloud
+backend and therefore anything doing perception on a Space.
+
 **And the endpoint is not answering, which is where this stands.** `turn.fastrtc.org` has no A
 record and `fastrtc.org` has no NS records at all, from three public resolvers and from the board
 — so the proxy both `fastrtc`'s own current code and `reachy_mini`'s #1182 point at cannot be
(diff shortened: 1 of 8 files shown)


pierre-rouanet:
Superseded. Main grew its own `spaces/vision-demo` and built it the other way round: 4e3506b replaced the WebRTC consumer with a receiver — the robot dials `wss://…/frames` and the Space decodes H.264 — and deleted `aiortc`, the DTLS cipher shim and, in its own words, "the ICE stack that could never complete from a data centre". That is this branch's `consumer.py` and `check_consumer.py`.

What this PR was carrying has landed or been answered elsewhere:

- **`filters.py`** is on main, same five filters.
- **The §6 measurement** — that a data centre cannot reach the robot without a relay — is confirmed there in stronger form: `offering relay candidates relays=5` across six sessions on olducky.
- **`turn.fastrtc.org` being dead** turned out to be the vanity alias only. The Space behind it never stopped answering, and main points at `https://fastrtc-turn-service.hf.space/credentials` (#258, #280). The recommendation here to report an outage to `fastrtc` is no longer right.
- **Two of the three publisher bugs** were fixed independently and better — `find -type f -o -type l` with `cp -L`, and `COPY . $HOME/app` instead of a hand-kept file list.

The third publisher bug is the one thing main still has, so it goes in on its own: #284.

Merging this as it stands would also revert `hf-robot-account` back to `updater::account` and delete §5.2 from the design doc. 168 commits behind, six files conflicting.

PR #226, opened 2026-09-07
robotd: refuse a non-socket path before bind, on every platform

### LINKED ISSUE #225: cargo test --workspace fails on a mac since #218: bind follows a dangling symlink there
since #218, `cargo test -p robotd` fails on a mac at `single_instance::non_socket_paths_are_not_removed` with "robotd did not exit", after the ten second wait. CONTRIBUTING says the whole workspace passes there with nothing excluded, and it did until that merge.

the test puts a dangling symlink at the socket path and expects robotd to refuse it. on linux it does, because `bind(2)` will not create through a trailing symlink and returns EADDRINUSE, which `claim_socket` turns into a refusal after seeing the path is not a socket. on macos `bind` follows the symlink and creates the socket at its target, so the daemon comes up serving through a path nobody asked for and the test waits for an exit that never comes. i checked the primitive on its own: bind over a dangling symlink whose directory exists succeeds on macos, and the target appears as a socket.

the regular file half of the same test is fine on both, since bind over an existing file is EADDRINUSE everywhere.

so `claim_socket` is relying on a linux kernel detail to enforce something it means on every platform, and the one platform the repo promises `cargo test --workspace` on is the one where it does not hold. the fix is to look at the path before bind rather than after, which is a few lines. branch coming.

on main at 5984efb, on a mac.

Nixxx19:
fixes #225

claim_socket checks what's at the path only after the bind fails, which relies on linux refusing to bind through a dangling symlink. macos follows it and makes the socket at the target, so robotd comes up serving a path nobody asked for and single_instance sits there waiting for a refusal that never comes.

checks the path before binding instead, so both refuse the same way.

### DIFF
--- a/robotd/src/main.rs
+++ b/robotd/src/main.rs
@@ -3155,18 +3155,28 @@ async fn claim_socket(socket_path: &Path) -> std::io::Result<(std::fs::File, Uni
     use std::os::unix::fs::{FileTypeExt, PermissionsExt};
 
     let lock = claim_lock(socket_path)?;
+    // Before bind, not after it. A regular file or a symlink at the path is a typo, and it is
+    // refused here on every platform rather than left to the kernel: Linux fails to bind over
+    // either, which is what made this look enforced, but macOS follows a dangling symlink and
+    // creates the socket at its target, so a daemon there came up serving through a path nobody
+    // asked for. That is the one platform `cargo test --workspace` is promised on.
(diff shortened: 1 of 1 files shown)


Nixxx19:
hi @pierre-rouanet, could you take a look at this one? cargo test currently fails on a mac and this fixes it.

pierre-rouanet (approved):

PR #230, opened 2026-09-07
the account token gets a group of its own, with mediad as its only member

### LINKED ISSUE #229: the account token is readable by every daemon in the robot group, not only mediad
`remote-access-design.md` §2.5 and §2.6 say the credential at `/etc/robot/hf-token` is `0640 root:robot`, so only root and `mediad` can read it, and `updater/src/account.rs` puts it in that group on purpose. the units say something else. `robot` is the socket group, it is how any client reaches a daemon's `0660` socket at all, so `btd.service`, `padd.service` and `tofd.service` all carry `SupplementaryGroups=robot`, and `install.sh` and `provision.sh` add the operator's own account to it as well. so the owner's hugging face bearer token, which §7 says can still write to their repositories, is readable by the bluetooth daemon that parses bytes from anyone in radio range, by the gamepad daemon that the repository describes as the process whose whole value is having no privileged access, by the depth sensor daemon, and by every shell user on the board.

none of those is a known way in, and this is a boundary drawn in the wrong place rather than a break. but the design spent a section on this file's mode precisely so that the boundary would be the file's, and the group it chose is the one that means everybody.

the fix is a group of its own with `mediad` as its only member, created by the `sysusers.d` files the release already ships, granted in `mediad.service`, and named in `account.rs`. i ran the two sysusers files through `systemd-sysusers` on a debian trixie root and got `robot-account:x:998:mediad`. branch coming.

on main at 5984efb.

Nixxx19:
fixes #229

the token is 0640 root:robot, and robot is the socket group — btd, padd, tofd and every operator with a shell are in it. so the hf token is readable by the gamepad daemon, which the repo describes as the process whose whole value is having no privileged access.

gives it a group of its own with mediad as the only member. the sysusers files create it and put mediad in, mediad.service grants it, and install.sh does the same for a board without systemd-sysusers.

a board that hasn't picked the group up yet does what it already did for a missing group: file stays root-only and the journal says so once. ran the two sysusers files through systemd-sysusers on trixie and got robot-account:x:998:mediad.


Nixxx19:
hi @pierre-rouanet, could you take a look at this one when you have a minute? the account token ends up readable by more daemons than it should be.

pierre-rouanet:
Thanks — the premise holds up. `robot` really is the socket group rather than a "robot daemons" group (`btd.service:53`, `padd.service:49`, `tofd.service:28`, `mediad.service:69`, plus the operator via `install.sh` and `provision.sh:350`), and `mediad/src/relay.rs:101` is the only reader of the token in the tree, so narrowing it costs nothing. The doc edits cover every place that said `root:robot` about the credential.

Parking it rather than merging it, because the packaging this changes is about to stop being the packaging we ship. The image is built with Yocto now, and that recipe installs the units verbatim while creating the accounts itself at build time — it deliberately does not install the `sysusers.d` files. So of the six files here, an image sees two: `mediad.service` and `account.rs`. Tracked as pollen-robotics/microduck_yocto#8, which has the one-line `GROUPADD_PARAM` change and has to land in the same commit as the `SRCREV_daemon` bump that first carries this.

Three things to fix when it comes back, and the first is the one that matters:

**The degradation is not the one the description claims.** "the file stays root-only and the journal says so once" is true of `hf_robot_account`'s writer, but this PR also adds the group to `SupplementaryGroups=` in `mediad.service`, and systemd refuses to start a unit whose supplementary groups do not resolve. `install.sh:474` says so in this repo's own words, which is why the `robot` group three lines below is a `die` and not a `warn`. So on a board where creation did not happen, the cost is not a token — it is mediad not starting: no camera, no WebRTC, no control channel. Make the `robot-account` failure a `die` too, matching `robot`. (Dropping the `SupplementaryGroups=` entry and relying on the `m` line would also work, since that setting extends rather than replaces the user's own memberships — but `mediad.service` is scrupulous about naming the groups it runs with, and I would keep that.)

**It does not reach the boards that have the problem.** The group is applied on save (`store.rs:save`), so an already-signed-in board updates, gets the group, restarts mediad, and keeps its existing `/etc/robot/hf-token` at `root:robot` until the next refresh write — which `maintain` starts about a week before a ~30-day expiry. Those are the only boards holding a token at all, so as it stands the hole closes for the fleet in up to three weeks. A `chgrp robot-account` and `chmod 0640` of the existing file, in `hooks/postinstall` after sysusers or on updaterd start, closes it at update. mediad re-reads on every connect attempt (`relay.rs:92-100`), so there is no restart to sequence.

**Smaller:** the `getent passwd mediad` guard in `install.sh` silently skips the `usermod` on a board without `mediad`, and the fallback loop right above it is `for daemon in btd padd` — mediad has no account fallback, so "the same fallback as the accounts" is not quite what happens. `remote-access-design.md:241` is now ~115 columns in a paragraph wrapped at ~100. And `the_token_group_is_not_the_socket_group` pins the negative but not the drift that costs something — the constant renamed and the unit left behind; `robotd-params/src/edit.rs:696` already `include_str!`s a shipped file in a test, and the same trick against `mediad.service` would make the copies of the string provably one.

Issue #220, opened 2026-09-07
Everything WebRTC carries should be timestamped at source


pierre-rouanet:
Nothing this transport carries is timestamped at source. A frame arrives when it arrives, a `robot.state` notification arrives when it arrives, and a consumer that wants to know *when* the robot saw or felt something has only its own clock — which, over a relay, measures the network rather than the robot.

That is fine for driving a robot you are watching, and wrong for most of what a remote consumer is interesting for. **SLAM is the case that makes it concrete**: monocular SLAM on a stream with no capture times runs; the moment somebody wants visual-inertial — the IMU this robot has, at 50 Hz, on the same control channel — the two series cannot be related except by guessing.

Design note: [`remote-webrtc.md` §11](docs/design/remote-webrtc.md).

## Two halves, and they are not the same problem

**Media.** RTP timestamps are relative to a random offset: they order frames and date none of them. The mechanism is the `abs-capture-time` RTP header extension, which carries a wall-clock capture time per packet. **First job is checking what actually surfaces it** — whether `webrtcsink` will negotiate it, and what a browser (`RTCRtpReceiver.getSynchronizationSources`, `getStats`) and `aiortc` expose of it. A header extension nothing on the receiving side surfaces buys nothing, and that determines whether this half is cheap or a fork of somebody's stack.

**The control channel.** Ours, and cheap: a monotonic reading plus the boot epoch that makes it comparable across processes, on every notification that describes a moment (`robot.state` first). The cost is a field per message and an argument about which clock — and the answer has to be the same clock the media path ends up dating frames with, or the two series still cannot be joined.

## Why it is its own change

It alters what every notification looks like, so it wants its own `API_VERSION` bump and its own decision rather than riding along with a transport. Deliberately kept out of the remote-access slices (#209, #219).

## Done looks like

- A consumer can take a decoded frame and a `robot.state` sample and say, with a bounded error it can state, which happened first.
- The error is bounded by something written down, not by the network.

Nixxx19:
took the first job on the media half from source, since it decides the shape of the rest.

[...]

PR #232, opened 2026-09-07
VSLAM telemetry on top of the MuJoCo twin: shared clock, head IMU, sensor poses, skeleton (API v24–v25)

apirrone:
## What this is

Everything a mapper running off the robot needs that `robot.state` and `tof.frame` did not carry,
plus the simulation it was developed against. Two layers, stacked:

1. **The MuJoCo twin** (`sim-remote-io`, 2,349 lines) — `robotd --sim`, `tofd --sim`,
   `mediad --sim-camera`, the `duck-ether` fake radio, `scripts/duck-sim`, and
   `docs/design/simulation.md`. Every one of these is behind a flag nobody passes on a robot.
   If `sim-remote-io` has its own PR, merge that first and this one shrinks to layer 2. If it does
   not, review it here as `docs/design/simulation.md` first, then `duck-control/src/sim.rs`.
2. **VSLAM telemetry** (~1,000 lines) — what the `microduck_vslam` repo (`duckslam`) consumes.

## Layer 2, by API version

**v24 — a clock every stream shares, and what a map needs from the robot**
- `t_ns` (`CLOCK_MONOTONIC`, ns) on `robot.state` and `tof.frame`, so a state sample and a depth
  frame from two daemons land on one axis without guessing the offset between their start times.
  `t` and `at_us` stay.
- `media.video` gains `mono_ns` / `real_ns`, the two clocks read at one instant, so RTP time
  (dated in wall-clock by RTCP sender reports) can be put on the same axis. This is a hook for
  #220, not a resolution of it: no `abs-capture-time`, and no error budget is written down yet.
- `robot.state.imu {gyro, quat}` — the trunk IMU as the loop reads it, not only projected gravity.
- `robot.state.frames {camera, tof, head_imu}` — sensor poses in the trunk frame at this tick's
  *measured* head joints, from the same head FK `robot.look` uses.
- `robot.model` — a new read: joint order, trunk height, ToF beam directions and FOV, the frames at
  head zero, and the skeleton topology. A client asks rather than transcribing the kinematics.

**Head IMU** — `tofd` reads the BMI088 on the HAT (same I²C bus as the ToF, its own thread) and
serves `head_imu.stream` / `head_imu.frame` (gyro, accel, Madgwick quat, temp) at 100 Hz. Routed
over WebRTC like `tof.stream`, refused on BLE. `--no-imu` turns it off. The `bmi088` crate comes
from `pollen-robotics/bmi088-rs` at tag `v0.1.1`.

**v25 — the whole skeleton's pose**
- `kinematics::Model::body_poses()` — whole-skeleton FK in one tree-order pass.
- `robot.state.skeleton` (15 trunk-frame poses per tick) + `robot.model.skeleton` (name, parent),
  so a viewer draws the real robot moving without its own copy of the kinematics.

**Camera intrinsics go through main's path, not a second one.** A short-lived `robot.model.camera`
duplicated what main already publishes in `media.video`, and is gone. Instead, the alpha family's
calibration (graphite, 2026-09-07, 80 ChArUco views, 0.84 px RMS at 1280×720) ships as
`robotd-params`' `[media.intrinsics]` default — the camera module and M12 lens are one part across
alpha, and the principal point alone sits 60 px off the centre the nominal record assumes. A
per-robot table still overrides. One serde trap fixed on the way: a field-level `default` was
silently beating the struct-level one with `None`; a test now pins that an absent key is the
family solve.

## Compatibility and cost

- Every field is additive (`serde(default)`; empty/absent from an older daemon), and `API_VERSION`
  goes 23 → 25. Deploy robot first, then consumers: old consumers ignore the new fields.
- The control loop assembles a state frame only while somebody is subscribed (unchanged guard);
  the skeleton FK adds microseconds to that. `robot.subscribe` stays refused over BLE, so the phone
  never sees the ~1.5 KB the skeleton adds per message.
- The one always-on cost on a robot is the head-IMU thread: ~4% of the 400 kHz I²C bus next to the
  ToF's ~5%, serialised by the kernel. On a board without the chip it retries with backoff up to 30 s.
- `tofd` now links `bmi088` → `nalgebra`/`simba`/`wide`: a bigger binary and a longer compile,
  nothing at runtime.

## Merged with main (0.11.0)

Conflicts resolved in the proto (Call count), `mediad` main (relay startup precedes the
`--sim-camera` source chain), `media.video` (clock pair rides inside main's `video_params`), and
`robotctl` (both `Enable` and `RebootMotors` arms). Two builds fixed: `RemoteIo` implements main's
new `RobotIo::reboot` (a simulated servo has nothing latched to clear, so it succeeds at once), and
the `robotctl` `RobotState` fixture carries `skeleton`.

## Verified

`cargo build/test/clippy --workspace --all-targets` clean under `-D warnings`; `duckslam` suite
78 passed against the new shapes. Not yet run since the merge: a dev-push to a robot — do that
before tagging, especially for the head IMU on a board with and without the HAT.

## Follow-ups (not in this PR)

- #220 proper: `abs-capture-time` investigation and a written error budget for frame↔state ordering.
- Loop closure in `duckslam` is not effective yet; this PR is the robot side only.


robotd/src/main.rs
```diff
@@ -2919,6 +2950,15 @@ async fn control_loop<T: RobotIo>(
                 },
                 theremin: theremin_state.clone(),
                 chorale: chorale_state.clone(),
+                t_ns: proto::clock::monotonic_ns(),
```
pierre-rouanet: `t_ns` is stamped here, at publish, but the sensors it labels were read at `safety.read()` — with `controller.step()` (the ONNX forward pass) in between. So every `robot.state` timestamp lags the joint angles it describes by one inference, systematically, every tick.

That is the exact error the shared clock is meant to remove: a mapper pairing a camera frame with `joints`/`frames` inherits the bias and no amount of clock sharing fixes it. Taking `monotonic_ns()` next to `safety.read()` and carrying it down costs nothing. `tofd` already does it this way — it stamps inside the read loop.

tof/src/main.rs
```diff
@@ -132,6 +134,26 @@ struct Args {
...
+    /// **The fake at the loop level, with a simulator behind it** — which is where a fake belongs
+    /// here: `sensor.rs` says in as many words that the off-board `Sensor` "is not a fake sensor and
+    /// must never become one", because the thing it stands for is a vendor C library talking to a
+    /// bus. A frame arriving from somewhere else is a different question from a sensor that lies.
+    ///
+    /// The simulator answers `{"op":"tof"}` with the same 8x8 of distances and per-zone statuses
+    /// this daemon publishes, so nothing downstream — `robotd`, `maploc`, the viewer — can tell.
+    /// Head-IMU (BMI088) sample rate, Hz. The chip's default bandwidth is 100 Hz.
+    #[arg(long, default_value_t = 100)]
+    imu_hz: u8,
```
pierre-rouanet, line 140: Merge artifact: the doc comment that starts "Read frames from a simulated body at `host:port`" ends with "Head-IMU (BMI088) sample rate, Hz" and is attached to `imu_hz`, while `sim` a few lines down has no doc comment at all.

So `tofd --help` prints three paragraphs about simulated depth as the help for `--imu-hz`, and nothing for `--sim`. (The swallowed text also mentions `maploc`, a name that appears nowhere else in this PR.)

tof/src/main.rs
```diff
@@ -503,31 +667,63 @@ async fn subscriber(
...
+                write_line(&mut write, &response).await?;
+                return stream_imu(&mut write, &mut imu_frames).await;
             }
             _ => {
                 let response = proto::Response::err(
                     id,
                     proto::Error::new(
                         proto::code::METHOD_NOT_FOUND,
-                        "tofd serves tof.stream and nothing else",
+                        "tofd serves tof.stream and imu.stream and nothing else",
```
pierre-rouanet: The method is `head_imu.stream`, not `imu.stream` — so this message names something a client cannot call. Same string in `updater/src/ipc.rs`, and the `ImuStatus` doc comment in `imu.rs` says "the `imu.stream` answer" too.

tof/src/imu.rs
```diff
@@ -0,0 +1,204 @@
...
+        let mut temp_c = 0.0f32;
+        while !shutdown.load(Ordering::Acquire) {
+            let tick = Instant::now();
+            let dt = (tick - last).as_secs_f32().clamp(1e-4, 0.2);
+            last = tick;
+            match ahrs.update(dt) {
+                Ok((gyro, quat)) => {
+                    let accel = ahrs
+                        .imu()
+                        .read_accelerometer_ms2()
```
pierre-rouanet: A gyro failure breaks out of the loop and reopens the chip; an accelerometer failure publishes `(0.0, 0.0, 0.0)` and carries on. Zero acceleration is indistinguishable from free-fall to a consumer, and nothing ever recovers from it — the stream just keeps serving zeros.

Same treatment as the gyro would be better: break, reopen, and let `status.lost` say why.

mediad/src/main.rs
```diff
@@ -122,6 +125,16 @@ struct Args {
...
 
+    /// Take frames from a duck in MuJoCo at `host:port` instead of a camera.
+    ///
+    /// The geometry has to match the simulator's camera — `--width 640 --height 360 --fps 15` for
+    /// its defaults — because the frames arrive raw and length-prefixed with no handshake, and a
+    /// mismatch is a picture nobody can read rather than an error the pipeline can recover from.
+    /// `mediad` says so and refuses the frame if the sizes disagree.
+    /// Takes precedence over `[media] camera`, which is a fact about a robot and not about this.
+    #[arg(long)]
+    sim_camera: Option<String>,
```
pierre-rouanet, line 137: `mediad` has no `--width`, `--height` or `--fps` — the frame geometry comes from `[media] quality`. Anyone following this line gets `error: unexpected argument '--width'`.

mediad/src/pipeline.rs
```diff
@@ -895,6 +903,97 @@ fn camera_source(camera: &Camera, fps: u32) -> Result<gst::Element> {
...
+    tracing::info!(%addr, "the simulated camera is feeding the pipeline");
+
+    loop {
+        reader.read_exact(&mut header)?;
+        let len = u32::from_le_bytes(header) as usize;
+        // A frame of the wrong size means the two ends disagree about the geometry, and pushing it
+        // would be a picture nobody can read. Said once, loudly, rather than a stream of noise.
+        if len != expected {
+            return Err(std::io::Error::other(format!(
+                "the simulator sent a {len}-byte frame and this pipeline is set up for {expected}                  — `--width`/`--height` must match the simulator's camera"
```
pierre-rouanet: Two things: the same `--width`/`--height` that `mediad` does not have (the geometry is `[media] quality`), and a run of about eighteen spaces inside the literal from a broken line wrap, which lands in the log as written.

mediad/src/pipeline.rs
```diff
@@ -895,6 +903,97 @@ fn camera_source(camera: &Camera, fps: u32) -> Result<gst::Element> {
...
+    let pushable = src.clone();
+    std::thread::Builder::new()
+        .name("sim-camera".into())
+        .spawn(move || {
+            let mut complained = false;
+            loop {
+                match read_frames(&addr, expected, &pushable) {
+                    Ok(()) => tracing::warn!(%addr, "the simulated camera closed"),
+                    Err(e) if !complained => {
+                        complained = true;
```
pierre-rouanet, line 951: `complained` is never cleared on a successful reconnect, so after the first failed connect every later connect failure is silent for the life of the process. `sim_loop` in `tofd` and `RemoteIo` both reset it on success.

robotd-params/src/lib.rs
```diff
@@ -371,9 +406,8 @@ impl Default for MediaParams {
             camera: true,
             quality: Quality::default(),
             bitrate: None,
-            // Nobody has measured this robot's camera. `mediad` says so on the wire rather than
-            // implying a measurement that did not happen.
-            intrinsics: None,
+            // The family's calibration, until this robot has its own — see the field.
+            intrinsics: Some(CameraIntrinsics::alpha()),
```
pierre-rouanet: With this default every alpha robot publishes `calibrated: true`, so the flag stops meaning "this robot was measured" and starts meaning "some alpha robot was". `camera.rs` built that distinction on purpose — "a consumer that needs better can then tell that it needs to ask" — and it is gone here with nothing in its place.

The trade itself is right; the lost signal is cheap to keep. A `source: "family" | "robot"` next to `calibrated`, or reserving `calibrated` for a per-robot solve, would do it.

Separately, worth one sanity check on the numbers: `cy = 471.9` on a 720-tall frame is 112 px below centre (`cx` is the 59 px the PR body describes). 0.84 px RMS says the solve is self-consistent, but an offset that large is worth confirming it was not solved on a rotated or cropped frame.

pierre-rouanet (approved):
Approving — the telemetry design holds up. Additive everywhere, the FK stays in `kinematics` so nothing downstream transcribes it, `Model::alpha()` is a cached static so the per-tick skeleton pass really is microseconds, and the state frame is still built only behind `receiver_count() > 0`. `body_poses()` is a clean one-pass tree walk.

Comments below are minor and none of them blocks the merge. The only one that is not text or naming is the `t_ns` stamp in `robotd`: it is taken at publish rather than at `safety.read()`, with the ONNX step in between, which is the systematic offset the field exists to remove. Worth a follow-up even if it lands as is.

One process note rather than a code one: there is no separate `sim-remote-io` PR open, so the twin (layer 1) is reviewed here or nowhere.

PR #234, opened 2026-09-07
configd: reading the pairing pin takes the authority that sets it

### LINKED ISSUE #233: the pairing pin can be read by any daemon on the socket, including the ones that must not hold it
`configd` gates every mutating call on the caller's uid and leaves reads open to anyone who can reach the socket, on purpose, so support can inspect a robot it may not change. `system.pairingPin` is counted as a read. it is not one. the pin is what authenticates a phone over ble, and once a phone holds it every mutating call on that transport is reachable, `net.connect`, `pad.pair`, `update.apply`, `account.login` with `force`. so a peer that may read the pin may change the robot by another route, and the gate on `system.setPairingPin` is only as strong as the gate on reading it back, which is none.

who can reach the socket is the `robot` group, and the units put every daemon in it: `btd`, `mediad`, `padd`, `tofd`. `btd` needs the pin, it answers the pairing exchange with it, and it is in `--allow-user` for that. the other three have no reason to hold it. `padd` is the one the repository describes as the process whose whole value is having no privileged access, and it parses input from whatever pad is bonded. today it can read the secret that turns a phone into an operator.

`app-path-design.md` says security over ble rests entirely on the pin being per robot. that is the right place to rest it, and it is why the pin should be readable by exactly the peers allowed to set it. a person reads it the way they set it, with sudo, which `system set-pin` already requires.

nothing fails today and nothing is known to be exploiting this. it is a secret sitting on the wrong side of a line the daemon draws carefully everywhere else. one line moves it, and the board test is unaffected since its `member` is in `--allow-user`. branch coming.

on main at 5984efb.

Nixxx19:
fixes #233

configd gates every mutating call on the caller's uid and leaves reads open on purpose. system.pairingPin counts as a read, but the pin is what a phone authenticates with over ble — whoever holds it can reach every mutating call that way.

and the socket group is every daemon, so padd and tofd can read it too.

so it now takes the same authority as setting it. btd keeps it, it's already in --allow-user for the pairing exchange, and a person reads it with sudo the same way they set it. every other read stays open.

if you'd rather a person read it without sudo, the other option is allowing any uid over 1000. went with the simpler rule since it matches what set-pin already needs.

### DIFF
--- a/btd/src/session.rs
+++ b/btd/src/session.rs
@@ -174,7 +174,7 @@ async fn dispatch(
             proto::code::PERMISSION_DENIED,
             format!(
                 "{} needs authentication first: send system.authenticate with the robot's PIN \
-                 (`robotctl system pin` on the robot)",
+                 (`sudo robotctl system pin` on the robot)",
                 call.method()
             ),
         );
--- a/configd/src/main.rs
+++ b/configd/src/main.rs
@@ -113,6 +113,20 @@ impl PeerPolicy {
     }
 }
 
+/// Whether a call takes the authority to change this robot.
+///
+/// Every mutating call does, and one read does with them: `system.pairingPin`. The PIN is not a
+/// status, it is a capability. Whoever holds it authenticates over BLE as the phone, and from there
+/// every mutating call is reachable, so a peer allowed to read it is a peer allowed to change the
+/// robot by another route. Reads are otherwise ungated on purpose, for [`PeerPolicy`]'s reasons,
+/// and the socket's group is every daemon that talks to one, the gamepad daemon included. A secret
+/// on that footing was readable by processes the design describes as having no privileged access.
+/// `btd` reads it to answer the pairing exchange and is in `--allow-user` for exactly that. A
+/// person reads it the way they set it, with `sudo robotctl system pin`.
+fn needs_authority(call: &proto::Call) -> bool {
+    call.is_mutating() || matches!(call, proto::Call::SystemPairingPin)
+}
+
 /// A user name to a uid.
 ///
 /// `SO_PEERCRED` reports a numeric uid, so a name has to become a number somewhere. Doing it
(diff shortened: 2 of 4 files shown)


Nixxx19:
hi @pierre-rouanet, mind a look at this one? the pairing pin can be read by any daemon on the socket, not just the one that needs it.

pierre-rouanet:
Thanks for this — the mechanics are right, and the test is in the right place. `btd` is the only daemon that actually reads the PIN (`btd/src/pairing.rs`), it is in `--allow-user btd` for exactly that, and `route.rs` already refuses `system.pairingPin` over BLE, so nothing on the phone path regresses. Reading it correctly as a capability rather than a status is a fair reading too.

I am closing it anyway, because of where the PIN stands today rather than anything wrong with the patch.

`docs/design/app-path-design.md` owns this mechanism, and §5.3 and §5.5 already say two things that decide it:

- the factory PIN is `000000` and public in this repository;
- `--require-pairing` is **off** by default, so the link is neither encrypted nor authenticated — "anyone in radio range during the exchange can read the PIN, and thereafter do anything a client may do".

So the hole this closes needs an attacker already running code on the robot as `padd` or `tofd`. The hole we have deliberately left open needs a radio and proximity. Closing the first while the second stands is the wrong order, and the cost is not free: `scripts/install.sh` puts the operator in group `robot`, so `robotctl system pin` works without sudo today, and this makes "what PIN do I type into the phone" a sudo step. First-connect friction is the thing we are least willing to spend right now.

What makes the PIN worth a uid gate is the per-robot provisioning obligation in `updater-design.md` §5.7 — something generating, printing and recording a PIN per robot. Once that exists and the link is encrypted, this gate is the obvious next step and your argument for it is the right one. Worth reopening then.

Two things for whoever picks it up at that point, since they are real misses here:

- The strings a person reads *at the moment they need the PIN* were not updated — `duckctl/src/main.rs` sends them to `robotctl system pin` in four places (the `--pin` help, and the three wrong-PIN / refused-call errors). All four would point at a command that now denies them. `docs/design/app-path-design.md` still says it too, and being the owning page, that is where the rationale belongs rather than only in a comment in `main.rs`.
- The refusal comes from `may_mutate`, which says "may not change this robot's configuration; add its user or group to `--allow-user`". For a read, that is confusing advice pointing at a unit file the operator should not be editing. It needs its own message.

PR #238, opened 2026-09-09
robotd: drop non-finite command targets instead of poisoning the command filters

hadelan:
Split out from https://github.com/pollen-robotics/microduck/pull/199.
JSON parses `1e400` as infinity, so a client can put one on the wire. Folded into the twist/head/body EMAs it is permanent (`ema += α·(inf − ema)` is `inf` on every later tick), and since the safety layer refuses non-finite joint targets rather than clamping them, a single bad `robot.move` froze the robot on its hold pose until reboot. The filters now drop non-finite targets.

Comes with a regression test. `cargo test -p robotd` and `cargo fmt --check` are green.

### DIFF
--- a/robotd/src/main.rs
+++ b/robotd/src/main.rs
@@ -1281,6 +1281,19 @@ impl Bringup {
     }
 }
 
+/// One low-pass step toward `target`.
+///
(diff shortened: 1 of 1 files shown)


pierre-rouanet (approved):
Thanks @hadelan !

PR #243, opened 2026-09-09
Adopt a factory-fresh servo as the one joint that stopped answering

### LINKED ISSUE #242: Replacing one motor at a time should auto flash with the right parameters
If a motor is broken, we should be able to just replace it with a new motor (baudrate 57600 and id 1) and the daemon should detect it, infer the right id and flash the appropriate parameters automatically (transparent to the user). 

It is able to infer the right id because of the missing motor on the bus, and id 1 is not used.

A subtlety is that we have to reboot the motor afterwards so that the shutdown error disapears.

apirrone:
Replacing a motor used to need a configuration tool: the new XL330 answers as ID 1 at 57 600 baud, the daemon sync_reads the fifteen expected IDs, and the register check failed forever with "is servo power on?" while the new servo sat invisible on the bus.

Now `open_bus` pings the expected IDs first. When exactly one is silent, it looks for ID 1 -- at 1 Mbps, then by reopening the port at 57 600 -- writes it the missing ID and then the bus's baud rate, reopens at 1 Mbps, runs the usual register check on it, and reboots it. The reboot clears the hardware-error alert the flash leaves set, which otherwise holds torque off until the battery is pulled; the alert register is read back and any remaining error is logged.

A complete bus pays fifteen pings for this and nothing else: the 57 600 probe only runs once a servo is known to be missing. Two missing servos are left alone, because there is no telling which one a fresh servo replaces. `robotd init` goes through the same open, so it accepts the new servo too.

Closes #242

Assisted-by: Claude Fable 5.1 <noreply@anthropic.com>

### DIFF
--- a/docs/design/robotd-design.md
+++ b/docs/design/robotd-design.md
@@ -273,6 +273,15 @@ saying so:
   swapped in arrives at 250, so the check is what removes a whole class of "why is it slow on
   this robot". `shutdown = 52` is the error mask that latches on overload, overheating and
   input-voltage faults.
+- **A swapped-in servo is adopted, not configured by hand.** A new XL330 answers as ID 1 at
+  57 600 baud, and neither is used on this bus. So before the register check, `open_bus` pings
+  the fifteen expected IDs; if *exactly one* is silent, it looks for ID 1 — first at 1 Mbps,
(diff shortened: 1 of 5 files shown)


apirrone:
Tested on a real robot, it worked !

PR #260, opened 2026-09-10
The servos report velocity and load every tick, and the wire dropped both

xaqmusic:
`bus.rs` reads one contiguous twelve-byte block per tick at register 124 — `present_pwm`, `present_current`, `present_velocity`, `present_position` — and unpacks all four into `Sensors`. Only position has ever reached `robot.state`. The other two are measured, paid for, and discarded at the socket.

Neither is recoverable from outside the daemon. Differencing `joints` between frames is not velocity: subscribers are decimated per connection, so a client at 10 Hz differences across five ticks, and any dropped frame becomes a spike. Load has no substitute at all — present current is the robot's only measure of external force: a foot taking weight, a hand pressing the beak, a servo on its way to an overload shutdown.

## The change

`RobotState` gains `velocities` (rad/s) and `currents_ma`, indexed as `JOINT_NAMES`, filled from the `Sensors` the loop already holds. Both ride as `#[serde(default, skip_serializing_if = "Vec::is_empty")]`, on the rule `odom` set: a frame from an older `robotd` still parses, and *absent* stays distinguishable from *zero* — a robot at rest and a robot that cannot say what it is doing must not render alike. `API_VERSION` goes 27 → 28 with an entry in the history, as v24 and v25 did for additive fields on the state stream.

`FakeIo` gains `set_velocities` / `set_currents_ma` beside `set_imu`. Positions are echoed from what was written, and deriving these the same way would make any test of the new path pass vacuously. `robotctl`'s monitor fixture names every field of the frame, so it gains the two, empty: the monitor draws neither.

## Cost

Measured on a frame assembled the way the loop assembles it, with the IMU, frames and skeleton present: 3042 B → 3346 B standing (+304 B, +10 %) and 3422 B walking (+380 B, +12 %); the spread is the digits a velocity carries once it is counts × 0.229 rpm. That is +15 to +19 KB/s at 50 Hz and +3 to +4 KB/s at a 10 Hz dashboard. The loop already skips frame assembly when nobody is subscribed, so a robot with no client pays nothing.

## Validation

Three tests: the blocks land under their documented names and survive a round trip; an older frame reports empty rather than zeros; and the control loop, driven by `FakeIo` with values distinguishable from each other and from the joint angles, puts both on the stream. duck-ipc-proto 50 → 52, robotd 150 → 151, the workspace 1363 → 1366, 0 failed. `cargo fmt --all --check` and `cargo clippy --workspace --all-targets` under `-D warnings` are clean, and `cargo llvm-cov` reports 72.17 % of lines on this branch against 72.11 % on `main` and the floor of 72.

What it does not show: everything here runs against `FakeIo`, so it covers the plumbing from `Sensors` to the wire and nothing below it. The unpack in `bus.rs` is untouched, and whether an XL330 reports a sensible current under load is a bench question — I have no hardware.

## Out of scope

`robotctl monitor` still displays neither field. And `serde_json` round-trips some f64 on this wire inexactly (`1.4000000000000001` comes back `1.4`), which affects `joints`, `targets` and `gravity` today and is about 1e-16 against a sensor resolution near 1e-3; the round-trip test compares approximately, on the rule `SafetyState` states when it drops `Eq`. Both belong in their own change, if you want them at all.

### DIFF
--- a/duck-control/src/io.rs
+++ b/duck-control/src/io.rs
@@ -256,6 +256,21 @@ impl FakeIo {
         self.sensors.imu = imu;
     }
 
+    /// Joint velocities the next [`RobotIo::read`] will report.
+    ///
+    /// Unlike positions these are never derived from what was written — a fake that echoed
+    /// its own targets back as velocity would make any test of the velocity path pass
+    /// vacuously.
+    pub fn set_velocities(&mut self, velocities: [f64; NUM_JOINTS]) {
+        self.sensors.velocities = velocities;
+    }
+
+    /// Present current per joint, mA. Same rule as [`Self::set_velocities`]: nothing derives
+    /// this, so a test has to state what the servos are supposed to be reporting.
+    pub fn set_currents_ma(&mut self, currents_ma: [f64; NUM_JOINTS]) {
+        self.sensors.currents_ma = currents_ma;
+    }
+
     pub fn positions(&self) -> [f64; NUM_JOINTS] {
         self.sensors.positions
     }
--- a/duck-ipc-proto/src/lib.rs
+++ b/duck-ipc-proto/src/lib.rs
@@ -326,7 +326,15 @@ pub const JSONRPC_VERSION: &str = "2.0";
 /// A new variant on a tagged enum is what a robotctl built before it cannot decode, which is the
 /// one reason this is a bump rather than a note: the tap is still `padd`'s own socket, and every
 /// other client is untouched.
-pub const API_VERSION: u32 = 27;
+///
+/// # v28 — measured joint velocity and load, on the state stream
+///
+/// [`RobotState::velocities`] and [`RobotState::currents_ma`]: the two blocks the control loop has
+/// read beside position on every tick and never published. Neither is recoverable from outside
+/// the daemon — differencing `joints` across a decimated subscription is not a velocity, and
+/// present current is the only measure of external force this robot has. Additive, on the rule
+/// `odom` set: absent from a daemon predating it, and absent stays distinguishable from zero.
+pub const API_VERSION: u32 = 28;
 
 /// The observation width every policy this robot family runs is built against.
 ///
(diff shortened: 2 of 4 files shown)


pierre-rouanet:
Reviewed, and I have pushed three commits to the branch rather than leaving a list — say if you would rather have had the list.

**The merge.** Seven API versions landed while this sat, so the entry is v35 and the constant goes 34 → 35. Nothing else conflicted. The description still says 27 → 28, and still describes the change as it was — worth a pass when you get a moment.

**The sign, and the fixture.** The `currents_ma` doc said the sign is dropped by the control loop; it is dropped in `bus.rs`, in the `.abs()` on the present-current word, and someone tracing a sign question would have read `main.rs` and found nothing. It also justified dropping it with "direction is inferable from `velocities`", which is false in exactly the three cases the same paragraph sells the field on — a joint holding a squat, a foot taking weight, a hand on the beak are all near-zero velocity at non-zero current, where a push and a pull read alike. Load is what a consumer wants, and that is reason enough on its own. `a_state()` arrived in this branch to stop a whole `RobotState` being spelled out per test, and then the two literals it was meant to replace were hand-patched with the new fields instead; both now override what they care about and take the rest from it.

**A zero that nothing measured.** This is the one that mattered. `Sensors` is fixed-size arrays, so a backend with nothing to report reports zeros — which is what a robot standing perfectly still under no load reports. `FakeIo` is that backend and is the default on any host that is not the robot, so `robotd --fake` put fifteen zero velocities and zero milliamps on the stream while `joints` swept through the home ramp: a dashboard draws a robot moving under no load rather than a robot that cannot say. The simulator had the same hole in `currents_ma`, which was optional in its frame and defaulted to zeros. The wire already draws the distinction the sample cannot, so `RobotIo::measures_velocity` / `measures_load` carry it out of the backend and the loop sends nothing instead. Both default true — a driver reading real servos has nothing to declare — and `Safety` forwards the two bools rather than handing out the IO it deliberately keeps private.

**`[control] publish_velocity_and_load`**, on by default. The way out if the bytes turn out to cost something on a robot, or a consumer turns out to mishandle the fields. Off means *absent*, which is what a daemon predating the fields sends, never a block of zeros.

So a block is empty for three reasons now — old daemon, nothing to measure, switched off — and a client cannot tell them apart, which is the point: all three say "this robot is not telling you". The field docs say so.

Workspace is green (`cargo test --workspace --all-targets`, 0 failed), `fmt --check` clean, and the one clippy warning on the branch (`manual_option_zip` in `chorale.rs`) is main's, not yours.

Your point about the unpack being untested below `Sensors` still stands — nothing here touches that, and whether an XL330 reports a sensible current under load is still a bench question.

pierre-rouanet:
Thanks @xaqmusic !

PR #262, opened 2026-09-10
systemd was not supervising the 3A engine at all

pierre-rouanet:
`rkaiq_3A.service` is a vendor unit, `Type=forking` over an init script whose `start_3A` is `rkaiq_3A_server 2>&1 | logger -t rkaiq &` — a pipeline, backgrounded, and then the script exits. systemd's forking heuristic finds no child it can claim, so on a board with the engine running and healthy:

```
$ systemctl is-active rkaiq_3A
active
$ systemctl show rkaiq_3A -p MainPID
MainPID=0
```

The processes do stay in the unit's cgroup, which is the one thing keeping `systemctl stop` honest — the final cgroup kill collects them, not the unit's own `ExecStop`. What that leaves broken:

- The unit reads `active` whether the engine is alive or dead, and nothing restarts it when it segfaults — the failure this platform actually has, the day the shim stops matching the kernel.
- The unit's `stop` is `killall rkaiq_3A_server`, so it reaches any engine on the box rather than the one it started.
- `reload` is stop-then-start inside the unit, so it never reaches the cgroup kill and leaks the previous run's `logger` every time.

## What the drop-in does now

`ExecStart` cleared and set to the engine itself, `ExecStop` cleared so the vendor `killall` never runs, and output to the journal under the unit instead of through a `logger` process — so `journalctl -u rkaiq_3A`, where this script's advice used to say `-t rkaiq`. systemd then knows the PID, kills the whole cgroup on stop, and can restart a dead engine.

`Type=simple` because the engine does not daemonise. The evidence is the `logger` itself: it was alive alongside the server on a board, which only holds if the server keeps the write end of that pipe open in the foreground. The script's report now checks `MainPID` after its own restart, so a board where that assumption turns out wrong says so where somebody is already looking.

## The cap is the interesting half

The drop-in's existing `ExecStartPost` bounces `mediad` on every start — the stream-start ordering invariant — and with `Type=simple` it fires as soon as the process is spawned, before a segfault at startup is known. Uncapped, a board with a mismatched shim would bounce the camera stream every few seconds for ever, which is worse than an engine that simply stays dead.

`StartLimitBurst=3` in three minutes bounds it: `mediad` is bounced three times, then systemd gives up and leaves the unit `failed` — at least a state `systemctl status` can report, where `active` with `MainPID=0` was not. `on-failure` rather than `always`, because the segfault exits non-zero and a clean exit from this engine has never been observed.

A failed engine is still invisible to `robotctl health`: `configd::units::MANAGED` is the seven units a release ships and `install.sh` manages, and this is a vendor unit we provision rather than ship. Now that it can reach `failed` at all, adding it is worth considering — a green picture has no reading on that report — but it widens what that list claims to cover, so it is not in this PR.

## Also

`setup-rkaiq.sh` joins CI's shellcheck list. It runs from `hooks/preinstall` on every update on every board and writes the drop-in that decides whether the engine is supervised; its sibling `setup-gstreamer.sh` has been linted there from the start. `setup-npu.sh` is the remaining unlinted installer and does not pass yet, so it is left out.

## Verification

`shellcheck --shell=sh`, `sh -n` and the `test -x` that CI's loop applies all pass, and the drop-in was rendered from the heredoc to confirm the generated unit file. The fix lands on a board through an ordinary update — the pre-install hook rewrites `robot.conf`, `daemon-reload` picks it up, and the script's closing `systemctl restart rkaiq_3A` makes the transition, killing the old untracked pair through the cgroup they are still in.

## Testing it on a board, without waiting for a release

The script needs `rkaiq-modinfo-shim.c` beside it or it refuses to run, so both files go over. From a checkout, without switching branches:

```bash
git fetch origin rkaiq-supervision
```

```bash
git show origin/rkaiq-supervision:scripts/setup-rkaiq.sh > /tmp/rk.sh
```

```bash
git show origin/rkaiq-supervision:scripts/rkaiq-modinfo-shim.c > /tmp/rk.c
```

```bash
scp /tmp/rk.sh "pierre@$(duckctl ip):setup-rkaiq.sh"
```

```bash
scp /tmp/rk.c "pierre@$(duckctl ip):rkaiq-modinfo-shim.c"
```

Then take a shell on the robot; everything below runs there, and `sudo` wants a password.

```bash
duckctl ssh
```

**The bug, before anything changes.** `Type=forking` with `MainPID=0` is the whole report, and both processes answer to init:

```bash
systemctl show rkaiq_3A -p Type -p MainPID -p NRestarts
```

```bash
ps -eo pid,ppid,comm | grep -E 'rkaiq|logger'
```

**Apply it.** Idempotent — the debs are installed and the shim is unchanged, so this rewrites the drop-in, reloads, restarts the engine and bounces `mediad` behind it. Its closing report says whether systemd is holding the engine:

```bash
sudo sh ~/setup-rkaiq.sh
```

**What should be different.** `Type=simple`, a real `MainPID`, no `logger` process at all, and the engine's own `DBG:` lines now under the unit ending in `wait stream start event success`:

```bash
systemctl show rkaiq_3A -p Type -p MainPID -p NRestarts
```

```bash
pgrep -a logger
```

```bash
journalctl -u rkaiq_3A -b --no-pager | tail -20
```

**That the engine still does its job** — video, and not green. From the laptop:

```bash
duckctl open
```

**That supervision now works**, which is the new behaviour and the thing worth actually testing. Kill the engine outright, wait about fifteen seconds for `RestartSec`, and ask again — `NRestarts=1` with a new PID, where before nothing would have noticed. The video drops for a moment, which is the `ExecStartPost` bouncing the stream on purpose:

```bash
sudo systemctl kill -s SIGKILL rkaiq_3A
```

```bash
systemctl show rkaiq_3A -p MainPID -p NRestarts
```

**That the cap holds.** Kill it four times inside three minutes and systemd stops trying rather than bouncing `mediad` for ever:

```bash
systemctl is-failed rkaiq_3A
```

```bash
sudo systemctl reset-failed rkaiq_3A
```

```bash
sudo systemctl start rkaiq_3A
```

**That a stop leaves nothing behind:**

```bash
sudo systemctl stop rkaiq_3A
```

```bash
pgrep -a rkaiq_3A_server
```

```bash
sudo systemctl start rkaiq_3A
```

To put the released behaviour back, run the release's own copy — it rewrites the drop-in from the version installed on the board:

```bash
sudo sh /opt/robot/daemon/current/scripts/setup-rkaiq.sh
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,11 +98,17 @@ jobs:
       # other script here, and was missing from this list while being the one file that runs on
       # every board on every update — the argument this comment opens with, applied to itself.
       #
+      # `setup-rkaiq.sh` runs from `hooks/preinstall` on every update too, and it writes a systemd
+      # drop-in that decides whether the 3A engine is supervised at all. Its sibling
+      # `setup-gstreamer.sh` has been linted here since the beginning; this one had not been.
+      # `setup-npu.sh` is the remaining unlinted installer and does not pass yet, so it is not
+      # here.
+      #
       # POSIX sh, not bash: the script is piped to `sh` by the documented one-liner, and
       # a bashism would work on a dev box and fail on the board.
       - name: Lint the installer
         run: |
-          for script in scripts/install.sh scripts/setup-board.sh scripts/setup-gstreamer.sh scripts/setup-login.sh scripts/setup-quiet-boot.sh scripts/migrate-network.sh scripts/provision.sh scripts/provision-board.sh scripts/ci-release-notes.sh scripts/robot-rescue scripts/dev-push.sh scripts/pad-link-test.sh scripts/pad-stack-report.sh scripts/seed-policies.sh; do
+          for script in scripts/install.sh scripts/setup-board.sh scripts/setup-gstreamer.sh scripts/setup-rkaiq.sh scripts/setup-login.sh scripts/setup-quiet-boot.sh scripts/migrate-network.sh scripts/provision.sh scripts/provision-board.sh scripts/ci-release-notes.sh scripts/robot-rescue scripts/dev-push.sh scripts/pad-link-test.sh scripts/pad-stack-report.sh scripts/seed-policies.sh; do
             sh -n "$script"
             shellcheck --shell=sh "$script"
             # The one-liner is only correct if the file is executable and self-contained.
--- a/docs/project/media-bringup.md
+++ b/docs/project/media-bringup.md
@@ -387,6 +387,58 @@ order is the whole trick:
 sudo systemctl stop mediad && sudo systemctl restart rkaiq_3A && sleep 2 && sudo systemctl start mediad
 ```
 
+### Nothing was supervising the engine
+
+The vendor unit is `Type=forking` over `/etc/init.d/rkaiq_3A.sh`, whose `start_3A` is
+
+```sh
+/usr/bin/rkaiq_3A_server 2>&1 | logger -t rkaiq &
+```
+
+— a pipeline, backgrounded, and then the script exits. systemd's forking heuristic expects one
+forked child and finds nothing it can claim, so on a board with the engine running:
+
+```
+$ systemctl is-active rkaiq_3A
+active
+$ systemctl show rkaiq_3A -p MainPID
+MainPID=0
+```
+
+Both processes are reparented to init, though they do stay in the unit's cgroup, which is the one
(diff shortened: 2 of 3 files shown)


Nixxx19:
hi @pierre-rouanet, was this parked or dropped? the branch is still there and main still has the vendor unit, so an engine that segfaults still reads as active with nothing to restart it.

happy to pick it up and finish it if you want it, just didn't want to reopen something you'd closed on purpose.

PR #263, opened 2026-09-10
tofd: only subscribe to the stream that was asked for

Nixxx19:
noticed tofd subscribes a connection to both streams before it even reads the request, so receiver_count is really just counting connections. ask for the imu and you're holding a depth receiver too.

moved the subscribe into the match arms so you only get the one you asked for.

nothing reads that count today so no behaviour change, but it'd be wrong for anything that did. test checks the counts over a real socket.


pierre-rouanet (approved):
Correct and narrow — the count this fixes is the one `mediad/src/detect.rs:201` and `robotd/src/main.rs:3089` already gate work on, so tofd was the odd one out even though nothing reads its count today.

One thing to fix in the body: "no behaviour change" is not quite true. Under the old code the receiver existed from accept, so a client could be handed frames buffered between connecting and asking. Now it starts at the subscribe. That is an improvement — a live depth stream wants current frames, not a stale burst — but it is a change, and the body should name it rather than claim none.

Nothing blocking. Happy for this to go in with the body amended.

PR #266, opened 2026-09-10
mediad: a websocket for programs, next to the datachannel for people

Nixxx19:
architecture.md §5.3 says an agent shouldn't have to go through webrtc for control: it doesn't want a 30fps h264 track to decode, it wants a frame every second or two and a state blob. and remote-webrtc.md §12 has the websocket surface on the deferred list as "a few dozen lines once §5's routing exists".

§5's routing exists now, so here it is. it's at /agent on the port the console is already served from.

turned out to be smaller than i expected because session::run already takes a pair of String channels and doesn't know what carries them. so the whole transport is pumping text frames into one and out of the other, and everything else is the code that was already there.

the part i think matters: it's the same api, not a second one. a line arriving on the socket goes through session::run against the same route table the datachannel uses, so what an agent can call is what a console peer can call, decided in one place. you can't add a method to one transport and not the other.

one thing that fell out: route::refusal said "not available over WebRTC", which is now wrong for half the callers. it says "over this transport" instead.

test opens a real websocket against the real router, checks hello comes back, checks a permitted call reaches robotd and a refused one is refused.

rebased on main: the console, /frame and /agent share the one listener now, and the snapshot test builds the router the same way.

cargo test -p mediad, clippy with -D warnings and fmt are green on linux. mediad is linux-only so a mac skips it.

### DIFF
--- a/docs/design/architecture.md
+++ b/docs/design/architecture.md
@@ -422,6 +422,11 @@ Same API behind all of them. "Run an LLM on a server that controls the robot"
 becomes: open a WebSocket, poll a frame, send intents — a few dozen lines, no
 media stack. That is what makes it genuinely easy.
 
+`mediad::agent` is that socket, at `/agent` on the port the console is already
+served from. Same API is not a promise here, it is the implementation: a line
+arriving there runs through `session::run` against `route`, which is the same
+dispatcher and the same table the datachannel uses.
+
 Note also that LLM latency (hundreds of ms to seconds) means the agent is a
 **high-level** controller: "go to the kitchen", "look at the person". Reactive
 control stays local in `robotd`. This is the correct split regardless of
--- a/docs/design/remote-webrtc.md
+++ b/docs/design/remote-webrtc.md
@@ -571,10 +571,12 @@ along with a transport. `remote-access-design.md` §9 carries it as open.
 
 ## 12. Deferred, with reasons
 
-- **A WebSocket surface for server-side programs** (`architecture.md` §5.3). Same JSON-RPC, no
-  media stack, `get_frame` returning a JPEG. It is a few dozen lines once §5's routing exists, and
-  it is what makes "an LLM drives the robot" easy — but it is a second transport and the first one
-  should work.
+- ~~**A WebSocket surface for server-side programs**~~ — built. `mediad::agent` serves it at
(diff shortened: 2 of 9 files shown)


pierre-rouanet:
Thanks. The code is well built: it reuses `session::run`, `route` and the per-peer `Pool`, there is no new port and no second table, and it reads `Media` the way the relay's control lane does. I'm not going to merge it as it stands, though, because I don't think the project needs this transport, at least not yet. Here's the reasoning, since #271 depends on it.

**§5.3 is out of date, and this PR takes it literally.** §5.3 was written before the rendezvous control lane existed. The case it names is an LLM *on a server*, which means off the LAN. `/agent` sits on the LAN console port, so a server can only reach it through a port forward, and `remote-webrtc.md` §4 calls that a deployment mistake. The off-LAN program is already covered:

- **Control:** `mediad::relay`'s control lane together with `spaces/shared/wire.py`. It's JSON-RPC over plain HTTP/SSE with no ICE, DTLS or `aiortc`. It works from a data centre, it's authenticated at both ends by the Hugging Face account, and it's answered by the same `session::run` and the same `route` table.
- **Frames:** WebRTC by default, with `media.stream` as the fallback (`faq.md`).

What `/agent` adds is a LAN script with no account. That's a real case, but a much narrower one than the PR describes. The roadmap also asks for the SDK's transport question to be settled "before writing either".

**If a LAN transport does come back later**, these are the things I'd want with it:

- **An `Origin` check.** A browser doesn't apply same-origin rules to WebSocket connections, so any page open on a LAN laptop could connect to `ws://<robot>:8080/agent` and call whatever `route` permits, `account.logout` included. Accepting only the console's own origin, or requests with no `Origin` header (what a script sends), closes that without costing a script anything.
- **A limit on notifications, like the relay lane's `Budget`.** At the moment a subscriber that stops reading stalls its own session on a full queue.
- **One doc owning it.** The mechanism goes in `remote-webrtc.md`, `architecture.md` §5.3 gets one sentence and a link, and `faq.md` says when to use this socket rather than WebRTC. §12 should drop the item instead of striking it through.

The follow-up on our side is to rewrite `architecture.md` §5.3 so it names the rendezvous lane as the server-side agent path. On #271 I've suggested building the SDK on that lane, which removes its dependency on this PR.

PR #271, opened 2026-09-12
sdk: a python client, and a loop to write behaviours in

Nixxx19:
stacked on #266, which is the socket this talks to. M5's finish line is "a server-side script can fetch a frame and send an intent in a few dozen lines", and this is the other half of it.

the investigation M5 says to do first, whether one client covers both webrtc and websocket, answers itself once you look at spaces/shared/control.py. that Rpc has already carried the same json-rpc over three transports: a datachannel through the rendezvous, a datachannel on the lan, and http with no webrtc at all. so the transport was never the hard part. the sdk binds a websocket to that same object rather than being a fourth implementation of the wire.

```python
from duck import Duck

with Duck("robot.local") as duck:
    print(duck.health()["healthy"])
    duck.move(vx=0.1)
```

the part worth having is watch(). one call each is fine for a script that asks a question, but a behaviour is a loop, and what it looks at arrives on three streams at three different rates: state at the control rate, depth at 15 hz, frames at whatever you asked for. merging those is what every caller would otherwise write for itself:

```python
for view in duck.watch(fps=4, camera=False):
    if view.nearest and view.nearest < 0.4:
        duck.move(vyaw=0.8)
    else:
        duck.move(vx=0.12)
```

view.nearest follows tof::Frame::zone rather than re-deriving the thresholds, including the bit that bites: the sensor returns negative distances on a failed convergence and still flags them valid, so taken at face value they are the nearest thing in the room. there's a test for exactly that.

no perception in here. view.frame is the jpeg as it arrived and what's in it is the caller's model to run.

on frames, i had this the wrong way round at first and fixed it: a still is a plain get to /frame on the console port, which is what a script wanting one picture should do, and webrtc is the path for anything actually consuming the camera. media.stream stays in as the fallback it is, frames only, no return path, for a long running stream where relay metering is the thing that matters. the readme says that now instead of arguing the outbound socket is the clever bit.

drove it against the real agent route with a fake robotd behind it: health, move and stop round trip, system.pairingPin comes back "not available over this transport", and the loop merges state and depth and picks the right nearest. ten tests on the wire and the depth rules, no robot needed.

rebased on main, so it sits next to the snapshot endpoint rather than on top of it.


pierre-rouanet:
Thanks, this is great and it's something we want. A Python SDK is the missing piece for anyone building against a duck, and this one is well designed. `with Duck(...) as duck: duck.move(vx=0.1)` is exactly the entry point it should have. `watch()` is the best part: merging state, depth and frames at the rate the caller chooses is what every behaviour would otherwise reimplement. Following `tof::Frame::zone` for `view.nearest` (negative distances included, with a test) is right, and so is reusing `Rpc` rather than writing a fourth wire. The examples read well too.

The one thing I'd change is what it sits on, for the reasons on #266: I don't want to build the SDK on `/agent`. `Duck("robot.local")` only works on the LAN, so it can't meet M5's "a server-side script" finish line. The path that already reaches a duck from anywhere is:

- **Control over `spaces/shared/wire.py`.** JSON-RPC through the rendezvous over plain HTTP/SSE, with no WebRTC, authenticated by the account and answered by the same `route` table. `Rpc` is already transport-agnostic, so this only changes `Duck.__init__`.
- **Finding the duck with `spaces/shared/rendezvous.py`** (`ducks(token)`), so the entry point is a robot's name plus a token, not a host.
- **Frames over WebRTC as an optional extra** (`pip install duck[video]` pulling in `aiortc`), so "send an intent, read state" stays light. `media.stream` stays available as the documented fallback.

Two smaller things in the frame path:

1. **The code doesn't match the description.** The description says stills come from `GET /frame` and streams from WebRTC. The code does neither: `watch()` opens a listening socket on the script's machine and has the robot dial back to it via `media.stream`. That needs the script to be reachable from the robot, which fails from behind NAT or from a data centre.
2. **`receive()` assumes every binary message is a JPEG** and discards the hello. `frames.encoding` can be `h264`, and `faq.md` says to branch on the hello (`decoder_for` in `spaces/vision-demo/receiver.py` does this).

On structure: `sdk/duck/rpc.py` is a copy of `spaces/shared/control.py`. Could the SDK take over `spaces/shared/` (`control.py`, `wire.py`, `rendezvous.py`), with the Spaces importing it, so there's one client rather than two copies that drift?

Two costs to state in the README: a LAN script now needs a signed-in account, and the rendezvous allows about 20 requests a second, which suits a behaviour loop at LLM speed and not teleop.

None of this touches the API you designed. It's a change underneath it, and afterwards the SDK no longer needs #266. Looking forward to landing it, and happy to help with the move to `wire.py` if that's useful.

PR #272, opened 2026-09-12
tof: reopen the sensor when it answers polls but never produces a frame

hadelan:
A wedged ranging engine still ACKs the I²C bus, so `data_ready` succeeds and answers `false` forever: `tofd` streams silence for the rest of the boot and nothing ever re-opens the sensor. The only "sensor is gone" signal today is a failed transaction, and this failure never produces one.

Put a clock on the answer: no frame for a few frame periods (2 s at the shipped 15 Hz, scaled so `--hz 1` gets the same margin in frames) breaks the poll loop into the existing backoff-and-reopen path. Armed at open, so a sensor that never produces a first frame is reopened on the same bound.

Regression test pins the watchdog scaling. `cargo test -p tof`, `clippy -D warnings`, `fmt --check` all green.

### DIFF
--- a/tof/src/main.rs
+++ b/tof/src/main.rs
@@ -259,6 +259,18 @@ fn quiet_period(hz: u8) -> Duration {
     Duration::from_secs_f64(1.0 / f64::from(hz.max(1))).saturating_sub(POLL_GUARD)
 }
 
+/// How long the poll may be told "no frame" before the sensor is called stuck.
+///
+/// This is the one failure `data_ready` cannot see: a sensor whose ranging engine has
+/// wedged still ACKs the bus, so the register read succeeds and the answer is `false`
+/// forever — without a clock on that answer the daemon streams silence for the rest of
+/// the boot and never re-opens the sensor. Scaled to the frame period so `--hz 1` gets
+/// the same margin in frames as the shipped 15 Hz gets in time.
+fn frame_watchdog(hz: u8) -> Duration {
+    let period = Duration::from_secs_f64(1.0 / f64::from(hz.max(1)));
+    Duration::from_secs(2).max(period * 4)
+}
+
 /// Bring the sensor up and stream from it, forever, with a backoff between
 /// attempts. Never returns until shutdown.
 fn sensor_loop(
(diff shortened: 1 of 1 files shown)


pierre-rouanet:
The diagnosis is right and the failure mode is real. Ranging is never gated in this daemon (`docs/project/tof-on-demand.md` closed that deliberately), so there is no legitimate multi-second gap in frames and no false-positive source for the watchdog. Arming at open also covers the "never produced a first frame" case, which the old code could not see either.

Two things I would raise before merge.

**1. The reopen loop is unbounded, and reopening is expensive.**

`open_sensor` → `Sensor::open` probes the device ID and uploads ~90 KB of ULD firmware, "a few seconds at 400 kHz" (`tof/src/sensor.rs:226`). On a permanently wedged sensor the cycle becomes: firmware upload (seconds) → 2 s of watchdog → `warn!` → `sleep(backoff)` → upload again, forever. It never escalates, because `backoff = RETRY_MIN` is reset on every *successful* open (`tof/src/main.rs:318`) and the open does succeed here — the failure is downstream of it. That is unlike the pre-existing paths, where a failing open climbs to `RETRY_MAX` = 60 s.

It matters because the bus is shared: the head IMU sits on the same `/dev/i2c-pihat`, so a stuck laser turns into a permanent firmware-upload flood contending with the IMU's two transactions per sample, plus a journal line every few seconds.

Cheapest fix is to not reset `backoff` when the previous attempt ended on the watchdog — let a sensor that keeps going silent back off toward 60 s, while a sensor that streams for a while and then dies still gets the fast retry.

**2. The status message now contradicts the diagnosis.**

The watchdog `break` falls into `status.down("the sensor stopped answering; retrying")` (`tof/src/main.rs:375`), but the whole premise is that the sensor *is* answering — that is why the watchdog had to exist. A subscriber asking why depth is unavailable gets a sentence pointing at the wrong failure. Carrying a reason out of the loop and into `status.down` is a few lines.

**Minor.** `max(2s, period * 4)` only leaves the 2 s floor at `hz == 1` (at `hz == 2` the two terms are exactly equal, and `hz` is a `u8`). So "scaled to the frame period" is in practice a special case for `--hz 1` — not wrong, just narrower than the comment reads. The test itself is good: it pins both ends and the `hz.max(1)` guard.

Nothing here is a correctness bug in the new code; (1) is the one I would want addressed.

pierre-rouanet (changes requested):
See comments above.

PR #283, opened 2026-09-14
updater: say when the update source last answered

### LINKED ISSUE #282: a robot that can't reach its update source looks up to date
if updaterd can't reach the update source (host blocked, dns broken, a clock tls won't accept), the scheduled check just fails into an info line in the journal and nothing else changes. update status and robotctl health still show the installed release and no error, so from the outside it looks exactly like a robot with nothing to install. it can sit like that for months and nobody would know.

nothing records when the source last actually answered, so there's no way to tell "no update" apart from "can't see updates".

Nixxx19:
fixes #282

updaterd now writes down when each component's source last answered with a manifest that verified, and update.status carries it as last_checked. robotctl health and update status show "source last answered 3 hours ago", and health warns once it's been quiet for a week, which is 28 missed checks at the shipped 6h interval.

only a real answer moves it: the source's latest, signed, for the right channel. a failed fetch, a manifest that doesn't verify and an apply --from a directory all leave it where it was, with a test for each. a clock before the preflight floor isn't recorded either, so a board that booted in 1970 doesn't come back reporting fifty years of silence.

it's its own small file in the state dir next to pins.json, and a damaged one reads as nothing instead of failing update.status. bumped the api to 29 for the new field.

one limit: a source replaying an old signed manifest still answers, so this catches a robot that can't reach its source, not one being fed stale metadata.

tests for updater, robotctl and duck-ipc-proto are green, and i broke the recording on purpose to make sure the new tests fail without it.


pierre-rouanet:
I force-pushed this branch (`df94b0f` → `077a95a`) after reviewing it — sorry for
writing to it directly rather than opening a PR against it. `df94b0f` is still
reachable by SHA, and the same three commits are on `pr283-review-fixes` in the
main repo if you want to compare.

**The rebase.** The branch was based on `61a746b`, and `main` has moved to
`API_VERSION = 34` since — v30–v34 added `homed`, `sitting`,
`ComponentStatus::degraded`/`reason`, `CpuThrottle` and `PolicySearchHit`. The
bump is 35 now and the note moved after the v34 block. One thing the rebase
surfaced: a `ComponentStatus` test helper in `robotctl` names every field, so
adding one to that struct needs `cargo check --workspace --all-targets` — a
scoped run does not see it.

**A source that has never answered.** `quiet_days` was `None` both for an
`updaterd` too old to carry the field and for one carrying it with nothing
recorded, and `quiet_source_warnings` dropped both. So a board blocked from its
source since it was provisioned printed no line and raised no warning, while a
board that answered once and then went quiet did — which is #282 exactly, with
this installed. The two silences are now told apart by the `hello` API version
that `collect_version_report` already has, against a new `proto::API_LAST_CHECKED`.

**A record ahead of the clock.** `(now - at).max(0)` read a future timestamp as
an age of zero. A board that checked with a fast clock, then had it corrected
backwards, then lost its source would sit at "fresh" for good — only a
successful check replaces the record, and by hypothesis there are none. It is
reported as unknown and warned about instead.

Both of those, plus "answered N ago" and "cannot say", are now one `SourceCheck`
decided in a single place. An `Option<i64>` could not carry it: absent meant two
opposite things.

**The warning followed `version` too.** It was added in
`collect_version_report`, which `robotctl version` shares, but only
`render_health` prints the line it refers to. It is `health`'s now.

**`health --json` carried the sentence.** `"9 days ago"` cannot be used by a
script and is stale the moment it is stored; `update status --json` already
answers in unix seconds, so this does too.

**An unreadable record is not a blank one.** `Checked::read_all` mapped every
failure to an empty map, so a transient IO error followed by a `record` wrote
back a map holding only the component being recorded, dropping the others. It
propagates now and the write is refused. Damaged JSON stays lenient — nothing in
it was readable anyway. The clock-floor skip also returned `Ok(())` silently, on
a board where every check succeeds and nothing is ever recorded; it logs at debug.

The design doc had option (3) as having no clock dependency. What is built
depends on the clock in both directions, so §8.4.2 now says which and how each
is handled.

Tests: robotctl 186, updater 228 + 84, duck-ipc-proto 51, all passing; fmt and
clippy clean apart from a pre-existing warning in `robotd/src/chorale.rs:545`.

Issue #294, opened 2026-09-15
Question: interpreting XL330 supply voltage in the battery shutdown policy


homalozoa:
I'm a hobbyist experimenting with the open-source software and XL330-M288 servos on my own setup. While debugging, I noticed a mismatch between the reported supply voltage and the battery shutdown threshold. I may be missing some context, so I would appreciate clarification.

## Observation

With automatic battery shutdown disabled for this debugging session, `robot.health` reported:

```json
"battery": {
  "volts": 4.98647771869289,
  "percent": 0.0
}
```

The control loop was running at about 50 Hz with zero consecutive bus errors. This is an observation from my personal setup, not a claim about official hardware.

## Source path I traced

In the public source at commit `fead66bb21195971fcbb2858a19b1e290a9f2031`:

- [`DynamixelIo::slow_sensors()`](https://github.com/pollen-robotics/microduck/blob/fead66bb21195971fcbb2858a19b1e290a9f2031/duck-control/src/bus.rs#L541) reads addresses 144–146 from the 15 `JOINT_IDS`, decodes the first two bytes with a factor of 0.1 V/count, and averages the positive voltage readings.
- [`publish_slow_sensors()`](https://github.com/pollen-robotics/microduck/blob/fead66bb21195971fcbb2858a19b1e290a9f2031/robotd/src/main.rs#L3250) smooths this value and stores it in `state.battery_v`, which is exposed through `robot.health`.
- [`BATTERY_EMPTY_V` / `BATTERY_FULL_V`](https://github.com/pollen-robotics/microduck/blob/fead66bb21195971fcbb2858a19b1e290a9f2031/duck-control/src/model.rs#L114) are 6.6 V / 8.2 V.
- The [automatic shutdown condition](https://github.com/pollen-robotics/microduck/blob/fead66bb21195971fcbb2858a19b1e290a9f2031/robotd/src/main.rs#L2322) uses `battery_empty_shutdown && battery_v > 0.0 && battery_v <= BATTERY_EMPTY_V`.

The [ROBOTIS XL330-M288 manual](https://emanual.robotis.com/docs/en/dxl/x/xl330-m288/) specifies an input range of 3.7–6.0 V, with 5.0 V recommended, and identifies address 144 as `Present Input Voltage`.

## Questions

1. What assumption makes the servo's `Present Input Voltage` a suitable estimate of battery voltage for this shutdown policy?
[...]

PR #295, opened 2026-09-16
duck-control: the shutdown mask clears input-voltage, and the span says which rail

Quchaosheng:
Follow-up to the question in #294, where a servo on a bench supply reporting ~5 V comes back as `percent: 0.0`. The conversion traced there is correct; the mismatch is that 6.6–8.2 V describes the *rail*, and the rail is the pack. Neither the constants nor the register mask said so, and chasing that down turned up a comment that says the opposite of what the code does.

**`shutdown = 52` is `0b110100`** — overload, electrical shock, overheating — with the input-voltage bit **clear**, where the factory's 53 sets it. That bit is what clears torque once the supply passes `Max Voltage Limit`, which nothing in this repo writes and which therefore stays at its default 7.0 V. A charged 2S pack sits above that, so the clear bit is what lets this bus run the pack's range across a servo rated to 3.7–6.0 V — and it is exactly the bit a reader would worry about when the pack is above the servo's own rating.

Both places that described the mask called it "an error mask that latches on overload, overheating and input-voltage faults", which is backwards. That is the doc-is-the-bug case `AGENTS.md` names, and it is a claim a reader cannot check without opening the manual.

**The battery span had the same gap from the other side.** It is pack-relative on purpose, but nothing said which rail it assumes, so a servo on a bench supply answering with its own 5 V is indistinguishable from a robot with a flat pack. §3.2 and the constants now name the rail, and say that a regulated servo rail would move these constants rather than reinterpret them.

The test pins the mask by its named bits, so a later edit that "restores the factory default" meets an assertion rather than a comment.

`cargo test -p duck-control` — 71 passed. `cargo fmt --all --check` and `RUSTFLAGS="-D warnings" cargo clippy -p duck-control --all-targets` clean.

Documentation and comments only; no behaviour changes, so no `API_VERSION` bump.

Assisted-by: Codex


Quchaosheng:
@pierre-rouanet the ci workflow is sitting at "1 workflow awaiting approval" here, so no job has run yet and there is no result to look at. Could you hit "Approve and run" when you get a minute?

Nothing else is blocking it: no conflicts with main, and the local checks were green — `cargo test -p duck-control` (71 passed), `cargo fmt --all --check`, `RUSTFLAGS="-D warnings" cargo clippy -p duck-control --all-targets`.

